PR #1, opened 2024-05-23
`Start/Stop` plugin

Keyrxng:
Resolves https://github.com/ubiquibot/plugins-wishlist/issues/9


Keyrxng:
image

src/types/plugin-input.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU extends SupportedEvents[T] = SupportedEvents[T]> {
+  stateId: string;
+  eventName: T;
+  eventPayload: TU["payload"];
+  settings: StartStopSettings;
+  authToken: string;
+  ref: string;
+}
+
+export const startStopSchema = T.Object({
```
gentlementlegen, line 13: Should have defaults

src/types/plugin-input.ts
```diff
@@ -0,0 +1,28 @@
...
+  disabledCommands: T.Array(T.String()),
```
gentlementlegen, line 14: Should have defined values for options (I guess "start", "stop")

src/types/env.ts
```diff
@@ -0,0 +1,11 @@
+import { Type as T } from "@sinclair/typebox";
+import { StaticDecode } from "@sinclair/typebox";
+import "dotenv/config";
+
+export const envSchema = T.Object({
+  GITHUB_TOKEN: T.String(),
```
gentlementlegen, line 6: Shouldn't it match `UBIQUIBOT_TOKEN`?

src/plugin-config.yml
gentlementlegen, line 1: Should not be here

.github/workflows/compute.yml
gentlementlegen, line 1: Since this is supposed to be a worker maybe we don't need this part?

Keyrxng: Makes sense then to also remove `main.ts` if it'll never be called in a workflow context and that is the workflow entry?

src/types/plugin-input.ts
Keyrxng: Aren't the defaults passed in via the kernel, I thought they were in `generateConfiguration()`, no? I will update

src/handlers/shared/structured-metadata.ts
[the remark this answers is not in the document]
Keyrxng: Because it was built into it I kept it there. Remove all reference to it from this plugin or?

So the new metadata system will be a standalone plugin added to a sequence or it will be n npm package called inside any plugin? a little unsure on that one

src/handlers/shared/stop.ts
```diff
@@ -0,0 +1,59 @@
+import { Context } from "../../types";
+import { addCommentToIssue, closePullRequestForAnIssue } from "../../utils/issue";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function stop(context: Context, issue: any, sender: { id: number; login: string }, repo: any) {
```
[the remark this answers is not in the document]
Keyrxng: a mistake is all, I had updated it in branched off commit and I've overlooked it starting from an earlier commit

src/types/plugin-input.ts
Keyrxng: I think just "start" makes sense.

It allows a user to stop a task if the command should be disabled but does not allow a new hunter to start, what do you think?

.cspell.json
```diff
@@ -1,10 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/streetsidesoftware/cspell/main/cspell.schema.json",
   "version": "0.2",
-  "ignorePaths": ["**/*.json", "**/*.css", "node_modules", "**/*.log"],
+  "ignorePaths": ["**/*.json", "**/*.css", "node_modules", "**/*.log", "./src/adapters/supabase/**/**.ts"],
   "useGitignore": true,
   "language": "en",
-  "words": ["dataurl", "devpool", "outdir", "servedir"],
+  "words": ["dataurl", "devpool", "outdir", "servedir", "Supabase", "supabase", "SUPABASE", "typebox", "ubiquibot", "signoff", "sonarjs", "mswjs"],
```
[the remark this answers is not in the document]
Keyrxng: You'd think so but nope those were thrown

src/types/plugin-input.ts
0x4007: Impossible for kernel to generate defaults for every possible plugin in the future.

.github/workflows/compute.yml
gentlementlegen: The workflow entry is the file you want as long as you specify it in the configuration.

src/handlers/shared/structured-metadata.ts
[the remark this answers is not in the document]
gentlementlegen: I do not know yet was it the best approach but supposedly if a plugin fail, as a GitHub Action they would dispatch an error and as a Worker they should return 500 to advise the kernel it failed. Problem if we handle that from the plugin itself that would mean every plugin should have to reimplement it.

src/adapters/supabase/helpers/logs.ts
```diff
@@ -0,0 +1,387 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
gentlementlegen, line 1: Can't this be achieved by `unknown` or a `<T>`?

src/adapters/supabase/helpers/user.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (error) throw console.error("Error getting location data", new Error(error.message));
+    return locationData;
+  }
+}
+
+const FAILED_TO_GET_USER = "Failed to get user";
+const SUCCESSFULLY_FETCHED_USER = "Successfully fetched user";
+
+interface LocationResponse {
```
gentlementlegen, line 83: I don't think this should be in the codebase

src/adapters/supabase/pretty-logs.ts
```diff
@@ -0,0 +1,190 @@
+import util from "util";
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
gentlementlegen, line 2: Possible to use `unknown` or `<T>`?

src/types/plugin-input.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const startStopSchema = T.Object({
+  disabledCommands: T.Array(T.String(), { default: ["start"] }),
+  labels: T.Object(
+    {
+      time: T.Array(T.String()),
+      priority: T.Array(T.String()),
+    },
+    {
+      default: {
```
gentlementlegen, line 23: Why are the labels relevant to the `start | stop`?

src/types/plugin-input.ts
Keyrxng: that makes sense

src/types/plugin-input.ts
Keyrxng: in `get-time-labels-assigned.ts` used to calc the `duration` part of the metadata and also in `analytics.ts` which is used in `get-multiplier-info.ts` which in part of the comment.

```ts
  const timeLabelsDefined = config.labels.time;
  ...
   const timeLabel = timeLabelsDefined.find((label: string) => label === _labelName);
    if (timeLabel) {
      timeLabelsAssigned.push(_label);
    }
```

src/adapters/supabase/helpers/user.ts
Keyrxng: I removed the literals and kept it in for now, does that make more sense until the `location` stuff is fully deprecated and removed?

src/handlers/shared/structured-metadata.ts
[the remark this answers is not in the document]
Keyrxng: As I was porting it over I was thinking that it would probably be best as a package

- callable anywhere
- useable by any plugin
- the only external call reliance is it's own calls to supabase
- best way to log more granular data without dispatching tens of workflows or fat bodied worker reqs or in the case of the kernel handling it just the fatal error reason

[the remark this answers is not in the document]
Keyrxng: Like `rpc-handler` and `configuration` it would be part of the eco-system of plugin dev tools/libs which is starting to grow

src/adapters/supabase/helpers/user.ts
gentlementlegen: If this is for the location why not using Supabase generated types?

Keyrxng: realised it's not in use at all anymore, my mistake, removed all together as the db types are already inferred where anything `location` related is concerned here

src/types/plugin-input.ts
gentlementlegen: It does make more sense but I don't think this would be a very common scenario. So in the end I would say let's have both there or none.

Keyrxng: understood, I've added stop as default as well

.github/workflows/worker-deploy.yml
```diff
@@ -0,0 +1,29 @@
...
+          wranglerVersion: "3.57.0"
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          secrets: |
+            SUPABASE_URL
+            SUPABASE_KEY
+            UBIQUIBOT_TOKEN
+        env:
+          SUPABASE_URL: ${{ secrets.SUPABASE_URL }}
+          SUPABASE_KEY: ${{ secrets.SUPABASE_KEY }}
+          UBIQUIBOT_TOKEN: ${{ secrets.UBIQUIBOT_TOKEN }}
```
rndquu, line 29: @Keyrxng `UBIQUIBOT_TOKEN` is the github app token, right? Isn't it supposed to be passed in a plugin invocation POST request (assuming we're running the plugin as a Cloudflare worker)?

rndquu:
@Keyrxng The plugin config example provided in the README throws `SyntaxError: "[object Object]" is not valid JSON` (I'm using [this](https://github.com/rndquu-org/test-repo/blob/159be92911c06a2585ee6088ac407b0748478dc5/.github/.ubiquibot-config.yml) one). Could you update README with a ready to use example?

src/types/plugin-input.ts
gentlementlegen: I get it but why should they be defined within the configuration I meant? They should just be parsed. Because the configuration from the other plugin responsible for the labels can also be changed. But you can be guaranteed to get something looking like `Time: <X DURATION` with X being a number and duration `hour | day | week`

Keyrxng:
> @Keyrxng The plugin config example provided in the README throws `SyntaxError: "[object Object]" is not valid JSON` (I'm using [this](https://github.com/rndquu-org/test-repo/blob/159be92911c06a2585ee6088ac407b0748478dc5/.github/.ubiquibot-config.yml) one). Could you update README with a ready to use example?

I probably made a mistake with it but I noticed when adjusting it that eslint was moving my comments out of the empty array.

@gentlementlegen I've brought over the template stuff, although there is no "Build" workflow, where does this come from?

src/types/plugin-input.ts
Keyrxng: I saw what you meant once I had another look at it so it's just parsing the labels from the issue itself now

src/handlers/shared/structured-metadata.ts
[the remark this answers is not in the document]
gentlementlegen: I think that:
- any fatal error should be handled by the kernel
- plugins should benefit from the new version of the `ubiquity-logged` that is capable of posting errors, and eventually can add metadata in the comment as well
- might not need a new package for this

gentlementlegen:
@Keyrxng When there is no `build` step I rather use `Knip` workflow to be done to trigger it.

.github/workflows/worker-deploy.yml
Keyrxng: I'll resolve this as @gentlementlegen has answered already pretty much

src/types/label.ts
```diff
@@ -0,0 +1,13 @@
+import { Static, Type } from "@sinclair/typebox";
+
+export const labelSchema = Type.Object({
```
gentlementlegen, line 3: Isn't this type provided within `Octokit`?

src/types/payload.ts
gentlementlegen, line 1: I feel like a lot of these types are also directly provided by `Octokit`, any reason to create them with TypeBox?

Keyrxng: I thought the same thing but also thought there must have been a reason for it, I had assumed that it had something to do with REST vs Webhook payloads or something along those lines.

I'll remove them and use the provided types since there doesn't appear to be a common knowledge reason why they were built like that

Keyrxng:
@gentlementlegen I noticed an error in `structuredMetadata.ts` for a regex using a char that's not supported until es2018 so that's why I've bumped that

src/types/payload.ts
```diff
@@ -1,236 +1,12 @@
...
-  repositories_added: Type.Optional(Type.Array(repositorySchema)),
-  changes: Type.Optional(changesSchema),
-});
-
+export type User = RestEndpointMethodTypes["users"]["getByUsername"]["response"]["data"];
+export type Issue = RestEndpointMethodTypes["issues"]["get"]["response"]["data"];
+export type Label = RestEndpointMethodTypes["issues"]["listLabelsOnIssue"]["response"]["data"][0];
 export type HandlerReturnValuesNoVoid = null | string | LogReturn;
+
+export const issueType = {
```
gentlementlegen, line 8: Should be an enum, and uppercase

Keyrxng: I've been of the belief that enums are considered bad TS practice in most cases. I've watched/listened to quite a bit of [Matt Pocock](https://www.youtube.com/@mattpocockuk) who seems to be an authority in the TS world afaik and he mentioned once (I'm paraphrasing here) that it was asked - if they could roll things back would they introduce enums and the answer was no they would have avoided it, this was someone who was involved in the development of TS I forget their name tho.

gentlementlegen: Nothing against this syntax, but The uppercase would be needed since it's used as a type imo.

Keyrxng: I thought eslint would complain if we made a follow this convention my mistake

Keyrxng: It's a debatable topic the whole `enum` and `const enum` thing isn't it. I was surprised to see them in use across UBQ at first actually I expected everyone to be in the against enum camp

gentlementlegen: I believe it is. I do not mind either, and I don't think the overhead is that big, but if it is mentioned in best practices for modern TS it makes sense to follow these rules. Feel free to open this topic inside of the `ts-template` repo.

src/adapters/supabase/helpers/user.ts
```diff
@@ -0,0 +1,79 @@
...
+  public async getLocationsFromRepo(repositoryId: number) {
+    const { data: locationData, error } = await this.supabase.from("locations").select("id").eq("repository_id", repositoryId);
+
+    if (error) throw console.error("Error getting location data", new Error(error.message));
+    return locationData;
+  }
+}
+
+const FAILED_TO_GET_USER = "Failed to get user";
+const SUCCESSFULLY_FETCHED_USER = "Successfully fetched user";
```
0x4007, line 79: This doesn't look right to declare the const at the end of the file does this actually work

README.md
```diff
@@ -1,32 +1,56 @@
...
 
-### Cypress
+To start a task, a hunter should use the `/start` command. This will assign them to the issue so long as the following is true:
 
-To test with Cypress Studio UI, run
+- Price labels are set on the issue
+- The issue is not already assigned
+- The hunter has not reached the maximum number of concurrent tasks
+- The command is not disabled at the repository or organization level
+- TODO: If the hunter meets the required karma and XP requirements
```
0x4007, line 15:
```suggestion
- TODO: If the hunter meets the required XP requirements
```
Don't introduce new terms. It will confuse future developers

src/adapters/supabase/helpers/user.ts
Keyrxng: No it doesn't that's my mistake, it works in `.cy` files tho I think that's been my confusion here

rndquu:
@Keyrxng Still getting the `SyntaxError: "[object Object]" is not valid JSON` error for [this](https://github.com/rndquu-org/test-repo/blob/a06641b2d350c23809654655572caa96e416c58f/.github/.ubiquibot-config.yml) config from README file. What am I doing wrong?

src/types/payload.ts
Keyrxng: @gentlementlegen eslint does complain it just must not have updated fast enough when I checked.

> 8:14  error  Variable name `IssueType` must match one of the following formats: strictCamelCase, UPPER_CASE

This err when running `yarn format`

Keyrxng:
@rndquu sorry about that you should be good to go now

src/types/payload.ts
0x4007: > Given the context of the discussion and the specifics of TypeScript usage, here is a response and possible resolution:
> 
> 1. **Enum vs Object with `as const`:**
>    - The preference for using objects with `as const` over enums aligns with the modern TypeScript best practices. It helps keep the codebase closer to standard JavaScript and avoids the additional generated code that enums introduce. This is particularly beneficial when considering tree-shaking and overall bundle size.
>    - **Supporting Reference:** As per the TypeScript documentation, "In modern TypeScript, you may not need an enum when an object with `as const` could suffice."
> 
> 2. **Naming Convention:**
>    - The issue regarding naming conventions is valid. The ESLint rule suggests that type-like entities should follow the UPPER_CASE naming convention.
>    - To resolve this, you should update the `issueType` to `ISSUE_TYPE` to satisfy the ESLint rules without switching to an enum.
> 
> ### Updated Code
> Here's how you can update the code to adhere to the naming conventions while sticking with the object with `as const`:
> 
> ```typescript
> import { LogReturn } from "../adapters/supabase/helpers/logs";
> import { RestEndpointMethodTypes } from "@octokit/rest";
> 
> export type User = RestEndpointMethodTypes["users"]["getByUsername"]["response"]["data"];
> export type Issue = RestEndpointMethodTypes["issues"]["get"]["response"]["data"];
> export type Label = RestEndpointMethodTypes["issues"]["listLabelsOnIssue"]["response"]["data"][0];
> export type HandlerReturnValuesNoVoid = null | string | LogReturn;
> 
> export const ISSUE_TYPE = {
>   OPEN: "open",
>   CLOSED: "closed",
> } as const;
> ```
> 
> ### Summary of Points:
> - **Enum vs Object with `as const`:** Use object with `as const` to stay in line with modern TypeScript practices and keep the code closer to standard JavaScript.
> - **Naming Convention:** Change `issueType` to `ISSUE_TYPE` to comply with the ESLint naming conventions and avoid naming errors.
> 
> This approach should address both the concerns about using enums and the ESLint naming conventions.

Keyrxng: So the outcome then is to use `const enum` and adjust the name yeah?

0x4007: Use the code it supplied thats fine.

src/adapters/supabase/helpers/user.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class User extends Super {
+  user_id: string | undefined;
+  comment_id: string | undefined;
+  issue_id: string | undefined;
+  repository_id: string | undefined;
+  node_id: string | undefined;
+  node_type: string | undefined;
```
gentlementlegen, line 16: I don't believe all of these are needed

src/utils/shared.ts
```diff
@@ -0,0 +1,32 @@
+import ms from "ms";
+import { Label } from "../types";
+
+export function calculateLabelValue(label: string): number {
+  const matches = label.match(/\d+/);
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) || 0 : 0;
+  if (label.toLowerCase().includes("priority")) return number;
+  // throw new Error(`Label ${label} is not a priority label`);
```
gentlementlegen, line 8:
```suggestion
```

src/adapters/supabase/helpers/user.ts
Keyrxng: you're right none of them are neither was one of the functions

src/utils/shared.ts
```diff
@@ -0,0 +1,32 @@
+import ms from "ms";
+import { Label } from "../types";
+
+export function calculateLabelValue(label: string): number {
+  const matches = label.match(/\d+/);
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) || 0 : 0;
+  if (label.toLowerCase().includes("priority")) return number;
```
0x4007, line 7: We should probably handle this in a new task but these "priority" and "time" labels are intended to be customizable so these shouldn't be hardcoded strings.

gentlementlegen: I actually wonder why is this needed for the `start | stop` commands? What influence do these tags have except that they have to be present on the issue in order to work? `calculateLabelValue` implies some calculation is done with these tags.

0x4007: Good point. This seems to be imported code from the v1 bot, which is later used to generate the price label. Perhaps this should be removed.

Keyrxng: The below metadata is directly tied to these functions when using the `/start` command in it's current state, taken from [here](https://github.com/ubiquity/devpool-directory-bounties/issues/29#issuecomment-2176741419)

```
<!-- Ubiquity - Assignment - start - 86f94c4
{
  "duration": 3600,
  "priceLabel": {
    "id": 5905388022,
    "node_id": "LA_kwDOJhaJks8AAAABX_0R9g",
    "url": "https://api.github.com/repos/ubiquity/devpool-directory-bounties/labels/Price:%2050%20USD",
    "name": "Price: 50 USD",
    "color": "1f883d",
    "default": false,
    "description": null
  }
}
-->
```

0x4007: I think that multiplier should be applied elsewhere, like on permit generation. Seems simple enough to aggregate all rewards and slap on the multiplier on the very end. 

Reason goes with multiplier, and is read only by `/query`

However, these are not essential functions because we barely used it.

src/adapters/supabase/helpers/user.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  async getWalletByUserId(userId: number, issueNumber: number) {
+    const { data, error } = await this.supabase.from("users").select("wallets(*)").eq("id", userId).single();
+    if ((error && !data) || !data.wallets?.address) {
+      console.error("No wallet address found", { userId, issueNumber }, true);
+      await addCommentToIssue(this.context, "```diff\n # Please set your wallet address with the /wallet command first and try again.\n```");
+      throw new Error("No wallet address found");
+    }
+
+    console.info("Successfully fetched wallet", { userId, address: data.wallets?.address });
```
gentlementlegen, line 20: All the `console` logs should be coming from the `logger` itself I believe.

Keyrxng: And we should use `Logs.ts` or the simple `context` logger until the `ubiquity/logger` is refactored?

Or should it be fired up like:

```ts
context.adapters = createAdapters(supabase, context);
context.logger = new Logs(supabase, 0, "ERROR", context)
```

gentlementlegen:
@rndquu That's true, but I think users cannot assign themselves without setting their wallet first, in our current workflow (bot v1)? Shall we change the workflow so users can assign themselves even without a wallet registered, and the wallet plugin would be in change to remind the user no wallet has been set?

0x4007:
> @rndquu That's true, but I think users cannot assign themselves without setting their wallet first, in our current workflow (bot v1)? Shall we change the workflow so users can assign themselves even without a wallet registered, and the wallet plugin would be in change to remind the user no wallet has been set?

I believe that the old version would allow you to self assign without a registered wallet. However, we should make this configurable. Default should not allow users to self assign unless their wallet is registered!

src/adapters/supabase/helpers/user.ts
gentlementlegen: @Keyrxng should be from the context.

Keyrxng:
Latest QA run:

- https://github.com/ubq-testing/command-start-stop/issues/1

- throws error if `startRequiresWallet` otherwise allows assignment but posts additional dedicated comment

rndquu (changes requested):
@Keyrxng Getting the following error on `yarn worker`:
```
✘ [ERROR] Could not resolve "../adapters/supabase/helpers/logs"

    src/utils/issue.ts:3:26:
      3 │ import { LogReturn } from "../adapters/supabase/helpers/logs";
```

Pls fix

src/types/context.ts
```diff
@@ -25,5 +26,5 @@ export interface Context<T extends SupportedEventsU = SupportedEventsU, TU exten
   adapters: ReturnType<typeof createAdapters>;
   config: StartStopSettings;
   env: Env;
-  logger: Logger;
+  logger: Logger | Logs;
```
gentlementlegen: I am not sure to understand this type, why could it be `Logger | Logs`?

Keyrxng:
I think you all were reviewing the commit before I removed the usage of `Logs`. I've used `PrettyLogs` for the logger and removed `Logs.ts`

- https://github.com/ubq-testing/command-start-stop/issues/2 New QA

image

.github/workflows/worker-deploy.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+      - uses: actions/checkout@v4
+      - uses: cloudflare/wrangler-action@v3
+        with:
+          wranglerVersion: "3.57.0"
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          secrets: |
```
rndquu, line 22: @gentlementlegen Do we need to pass `CLOUDFLARE_ACCOUNT_ID` here ([related PR](https://github.com/ubiquity/ubiquibot-kernel/pull/63))?

src/adapters/supabase/helpers/user.ts
```diff
@@ -0,0 +1,63 @@
...
+        await addCommentToIssue(this.context, "```diff\n# Please set your wallet address with the /wallet command in order to receive a task reward.\n```");
+      }
+    } else {
+      this.context.logger.info("Successfully fetched wallet", { userId, address: data.wallets?.address });
+    }
+
+    return data?.wallets?.address || null;
+  }
+
+  public async getMultiplier(userId: number, repositoryId: number) {
```
rndquu, line 29: @Keyrxng As far as I understand `getMultiplier()`, `_getAccessData()` and `getLocationsFromRepo()` are not used anywhere. If so pls remove them.

Keyrxng: my mistake, resolved.

.github/workflows/worker-deploy.yml
gentlementlegen: Not required, but safer so yes it would be nice to add it here.

Keyrxng: @rndquu done

rndquu (approved):
Works fine

Moved this https://github.com/ubiquibot/command-start-stop/issues/4 to a separate issue since this PR is already too big

.github/workflows/worker-deploy.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: Deploy
+    steps:
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
```
gentlementlegen, line 16:
```suggestion
          node-version: "20.10.0"
```

README.md
```diff
@@ -1,32 +1,56 @@
-# `@ubiquity/ts-template`
+# `@ubiquibot/Command-Start-Stop`
```
gentlementlegen, line 1:
```suggestion
# `@ubiquibot/command-start-stop`
```

package.json
```diff
@@ -76,5 +83,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
+  },
+  "packageManager": "yarn@4.2.2"
```
gentlementlegen, line 87: Are we really using `yarn 4` for this repo? Is there no side effect for this?

rndquu: Works fine with yarn v4, at least in the "worker mode". As far I understand, [corepack](https://yarnpkg.com/corepack) (binary installed with the NodeJS) makes sure that yarn version from the `packageManager` field is used within the project.

gentlementlegen: As long as it is supported within Cloudflare then we are good to go I guess!

tests/main.test.ts
```diff
@@ -65,18 +66,33 @@ describe("User start/stop", () => {
...
-    expect(errorSpy).toHaveBeenCalledWith("You are not assigned to this task");
+    expect(output).toEqual({ output: "You are not assigned to this task" });
+  });
+
+  test("User can't stop an issue without assignees", async () => {
+    // using the second issue
+    const issue = db.issue.findFirst({ where: { id: { equals: 6 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 1 } } }) as unknown as Sender;
+
+    console.log(issue.assignees, issue.assignee?.login, sender)
```
gentlementlegen, line 88: Do we want these logs?

tests/main.test.ts
```diff
@@ -1,24 +1,475 @@
...
+
+    expect(output).toEqual({ output: "You are not assigned to this task" });
+  });
+
+  test("User can't stop an issue without assignees", async () => {
+    // using the second issue
+    const issue = db.issue.findFirst({ where: { id: { equals: 6 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 1 } } }) as unknown as Sender;
+
+    console.log(issue.assignees, issue.assignee?.login, sender)
```
gentlementlegen, line 88:
```suggestion
```

tests/main.test.ts
```diff
@@ -1,24 +1,475 @@
...
+
+    expect(output).toEqual("Task unassigned successfully");
+  });
+
+  test("User can't stop an issue they're not assigned to", async () => {
+    // using the second issue
+    const issue = db.issue.findFirst({ where: { id: { equals: 2 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 1 } } }) as unknown as Sender;
+
+    console.log(issue.assignees, issue.assignee?.login, sender)
```
gentlementlegen, line 73:
```suggestion
```

README.md
```diff
@@ -1,32 +1,56 @@
...
 
-## Testing
+### Start a task
 
-### Cypress
+To start a task, a hunter should use the `/start` command. This will assign them to the issue so long as the following is true:
 
-To test with Cypress Studio UI, run
+- Price labels are set on the issue
+- The issue is not already assigned
```
0x4007, line 12: This reminds me that it would be really interesting in the future to support teams. Where I could write on a new task 

```
/start @Keyrxng 
```

And we would both be assigned (and be eligible to split the reward.)

I know a lot of developers work as lone rangers but it could make the experience more collaborative and exciting if we have a massive tasks marketplace and teams hunt projects to co-op on together.

README.md
```diff
@@ -1,32 +1,56 @@
...
-yarn cy:run
+#### Note: The command name is `"start"` when configuring your `.ubiquibot-config.yml` file.
+
+To configure your Ubiquibot to run this plugin, add the following to the `.ubiquibot-config.yml` file in your organization configuration repository.
+
+```yml
+- plugin: http://localhost:4000 # or the URL where the plugin is hosted
+  name: start-stop
+  id: start-stop-command
+  description: "Allows a user to start/stop a bounty without negative XP impact"
```
0x4007, line 35:
```suggestion
  description: "Allows a user to start/stop a task without negative XP impact"
```

We completely stopped using "bounty" or "bounties" because this is a crypto-native term and our aspiration is to work with non crypto developers as soon as possible.

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "ts-template",
+  "name": "@ubiquibot/command-start-stop",
   "version": "1.0.0",
-  "description": "Template repository with TypeScript support.",
+  "description": "Enables the assignment of tasks to hunters and the graceful unassignment.",
```
0x4007, line 4:
```suggestion
  "description": "Enables the assignment of tasks to contributors and the graceful unassignment.",
```

No more bounty hunter vernacular.

src/adapters/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Database } from "./supabase/types/database";
+
+export function createAdapters(supabaseClient: SupabaseClient<Database>, context: Context) {
+  return {
+    supabase: {
+      user: new User(supabaseClient, context),
+      super: new Super(supabaseClient, context),
+    },
+  };
+}
```
0x4007, line 14: I think we should use each organization's own data stores. 

In exchange for having a universal/system wide registration system, we receive the following benefits:

1. decentralized data storage (within each org)
2. each org can edit the values as they see fit, providing them full control over their org and bot behaviors within. 
3. not needing to manage/maintain a supabase database (which can adversely affect the services of all of our users)
4. easier maintenance on the plugins, such as this, for not having to spin up a supabase.

src/handlers/shared/check-task-stale.ts
```diff
@@ -0,0 +1,12 @@
...
+  let staleToDays: number | undefined;
+  let isTaskStale = false;
+
+  if (staleTask !== 0) {
+    days = Math.floor((new Date().getTime() - new Date(createdAt).getTime()) / (1000 * 60 * 60 * 24));
+    staleToDays = Math.floor(staleTask / (1000 * 60 * 60 * 24));
+    isTaskStale = days >= staleToDays;
+  }
+  return isTaskStale;
+}
```
0x4007, line 12: Doesn't make sense to allow `| undefined`

Just declare it inside of the if block if thats the only place its used.

src/handlers/shared/generate-assignment-comment.ts
```diff
@@ -0,0 +1,45 @@
...
+  timeZone: "UTC",
+  timeZoneName: "short",
+};
+
+export async function generateAssignmentComment(
+  context: Context,
+  issueCreatedAt: string,
+  issueNumber: number,
+  senderId: number,
+  duration: number | null = null
```
0x4007, line 18: You could also use `0` for duration to keep it as a `number`?

src/handlers/shared/start.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
+    await addCommentToIssue(context, "```diff\n! Too many assigned issues, you have reached your max limit.\n```");
+    throw new Error(`Too many assigned issues, you have reached your max limit of ${maxConcurrentTasks} issues.`);
+  }
+
+  // is it assignable?
+
+  if (issue.state === ISSUE_TYPE.CLOSED) {
+    await addCommentToIssue(context, "```diff\n! The issue is closed. Please choose another unassigned bounty.\n```");
```
0x4007, line 55:
```suggestion
    await addCommentToIssue(context, "```diff\n! The issue is closed. Please choose another unassigned task.\n```");
```

Get rid of all "bounty" references

src/handlers/shared/stop.ts
```diff
@@ -0,0 +1,56 @@
...
+export async function stop(context: Context, issue: Context["payload"]["issue"], sender: Context["payload"]["sender"], repo: Context["payload"]["repository"]) {
+  const { logger } = context;
+  const issueNumber = issue.number;
+  const out = { output: null };
+
+  // is it an issue?
+  if (!issue) {
+    logger.info(`Skipping '/stop' because of no issue instance`);
+    return out;
+  }
```
0x4007, line 13:
```suggestion
```

We can probably get rid of these checks. I know they were necessary in the old bot because of the poor type management.

src/handlers/shared/stop.ts
```diff
@@ -0,0 +1,56 @@
...
+    issue_number: issueNumber,
+    assignees: [sender.login],
+  });
+
+  logger.info("You have been unassigned from the task", {
+    issueNumber,
+    user: sender.login,
+  });
+
+  addCommentToIssue(context, "````diff\n+ You have been unassigned from this task.\n````").catch(logger.error);
```
0x4007, line 54:
```suggestion
  addCommentToIssue(context, "```diff\n+ You have been unassigned from this task.\n```").catch(logger.error);
```

Four backticks is strange.

src/handlers/shared/structured-metadata.ts
```diff
@@ -0,0 +1,35 @@
...
+    // otherwise we want to hide it
+    metadataSerialized = metadataSerializedHidden;
+  }
+
+  return metadataSerialized;
+}
+
+export default {
+  create: createStructuredMetadata,
+};
```
0x4007, line 35: I wonder if we should proceed with the metadata system. 

It seems  really nice to have but we need to make it way more convenient for developers to work with (via our SDK) for this to catch on and be useful.

src/handlers/user-start-stop.ts
```diff
@@ -0,0 +1,25 @@
...
+  const directive = comment.body.split(" ")[0].replace("/", "");
+  const { disabledCommands } = config;
+  const isCommandDisabled = disabledCommands.some((command: string) => command === directive);
+
+  if (isCommandDisabled) {
+    await addCommentToIssue(context, "```diff\n! The /start command is disabled for this repository.\n```");
+    throw new Error(`The '/${directive}' command is disabled for this repository.`);
+  }
+
+  if (directive === "stop") {
```
0x4007, line 18: Why are these called directives? They should be called commands.

src/main.ts
```diff
@@ -0,0 +1,48 @@
...
+  const inputs: PluginInputs = {
+    stateId: payload.stateId,
+    eventName: payload.eventName,
+    eventPayload: JSON.parse(payload.eventPayload),
+    settings,
+    authToken: payload.authToken,
+    ref: payload.ref,
+  };
+
+  await startStopBounty(inputs, env);
```
0x4007, line 23:
```suggestion
  await startStopTask(inputs, env);
```

No bounty

src/main.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return returnDataToKernel(inputs.authToken, inputs.stateId, {});
```
0x4007, line 25:
```suggestion
  return toKernel(inputs.authToken, inputs.stateId, {});
```

Perhaps this is more succinct and still clear.

src/plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+  const supabase = createClient<Database>(env.SUPABASE_URL, env.SUPABASE_KEY);
+
+  const context: Context = {
+    eventName: inputs.eventName,
+    payload: inputs.eventPayload,
+    config: inputs.settings,
+    octokit,
+    env,
+    logger: new PrettyLogs(),
+    adapters: {} as ReturnType<typeof createAdapters>,
```
0x4007, line 22:
```suggestion
    adapters: createAdapters(supabase, context);
```

Doesn't this work?

src/types/plugin-input.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU extends SupportedEvents[T] = SupportedEvents[T]> {
+  stateId: string;
+  eventName: T;
+  eventPayload: TU["payload"];
+  settings: StartStopSettings;
+  authToken: string;
+  ref: string;
+}
+
+const oneDay = 24 * 60 * 60 * 1000;
```
0x4007, line 14:
```suggestion
const ONE_DAY = 24 * 60 * 60 * 1000;
```

src/utils/get-linked-prs.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+interface GetLinkedResults {
+  organization: string;
+  repository: string;
+  number: number;
+  href: string;
+}
+
+export async function getLinkedPullRequests(context: Context, { owner, repository, issue }: GetLinkedParams): Promise<GetLinkedResults[]> {
```
0x4007, line 17: I hope this works. It worked for me but @gentlementlegen had problems in the past when using it in some tests.

src/utils/issue.ts
```diff
@@ -0,0 +1,171 @@
...
+  const payload = context.payload;
+
+  try {
+    return await context.octokit.paginate(
+      context.octokit.issues.listForRepo,
+      {
+        owner: payload.repository.owner.login,
+        repo: payload.repository.name,
+        state: ISSUE_TYPE.OPEN,
+        per_page: 1000,
```
0x4007, line 20: I wonder if `1000` is the correct number. Why did you use this?

src/utils/issue.ts
```diff
@@ -0,0 +1,171 @@
...
+      repo: payload.repository.name,
+      issue_number: issueNumber,
+      body: comment,
+    });
+  } catch (e: unknown) {
+    context.logger.fatal("Adding a comment failed!", e);
+  }
+}
+
+//// Pull Requests \\\\
```
0x4007, line 48:
```suggestion
// Pull requests
```

src/utils/issue.ts
```diff
@@ -0,0 +1,171 @@
...
+  const { repository } = context.payload;
+  try {
+    await context.octokit.rest.pulls.update({
+      owner: repository.owner.login,
+      repo: repository.name,
+      pull_number: pullNumber,
+      state: "closed",
+    });
+  } catch (err: unknown) {
+    context.logger.fatal("Closing pull requests failed!", err);
```
0x4007, line 60: I was thinking of saving fatal for emergencies that can be triaged by our team.

For example, a fatal error in the kernel occurs that can affect every partner and user of our system. So as soon as it occurs, it should ping our Telegram bot, which should relay the message in a triage group on Telegram. 

That way our incident response time can be minimal.

Obviously this should only be saved for actual, fatal emergencies. 

I don't think this qualifies.

src/utils/issue.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+
+export async function closePullRequestForAnIssue(context: Context, issueNumber: number, repository: Context["payload"]["repository"]) {
+  const logger = context.logger;
+  if (!issueNumber) {
+    throw logger.fatal("Issue is not defined");
```
0x4007, line 67: Same here regarding fatal.

src/utils/issue.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  if (!linkedPullRequests.length) {
+    return logger.info(`No linked pull requests to close`);
+  }
+
+  logger.info(`Opened prs`, linkedPullRequests);
+  let comment = `These linked pull requests are closed: `;
+  for (let i = 0; i < linkedPullRequests.length; i++) {
+    await closePullRequest(context, linkedPullRequests[i].number);
+    comment += ` <a href="${linkedPullRequests[i].href}">#${linkedPullRequests[i].number}</a> `;
```
0x4007, line 84:
```suggestion
    comment += ` ${linkedPullRequests[i].href} `;
```

Not sure if this is better, but I remember seeing this render in a diff block so there's no need to dress it up if formatting isn't supported.

src/adapters/index.ts
Keyrxng: I've branch from the latest commit of this PR and threw something together quickly but I have a question

I'm assuming there would be a dedicated `ubiquibot-plugin-storage` repo the partner would create which would be a centralized place which plugins for that install could use?

The incoming payload has the repo in which the `/start` command was used so I can't read/write to every repo and create a `start-stop-storage.json` but I could from a dedicated repo easily. 

The impl has pulled gh ids and wallets from our sb and has written to `Record<id, wallet>` then all sb logic removed and replaced with read/write/commit to repo but without the above being true, where exactly would each plugin store the data in the partner's org?

<hr/>

I think it's a good idea because any installed plugins by a partner should be trusted so each plugin having the ability to easily access the storage of other installed plugins makes things really easy for plugin DX and interoperability.

Even if the storage repo is private just knowing the storage structure of various plugins makes it easier to build on top of things. E.g. you could name the storage for this simply `wallets.json` and then other plugins need to confirm a registered user for that org checks `wallets.json`, further, `/wallet` could read/write to the same file separating logic but sharing storage

src/handlers/shared/structured-metadata.ts
Keyrxng: I agree it should be integrated to the SDK to be more beneficial

It would be very effective as an error wrapper/fallback of sorts. Called in catch blocks etc and before the end of plugin execution it handles posting/storage/notifying making using it a breeze

With the above actually you could hoover up metadata throughout the whole plugin flow which could be v useful (logs would do this but it would save hunting for them)

src/plugin.ts
Keyrxng: No because `context` is not instantiated at this point and you cannot use a `this` context either

src/utils/get-linked-prs.ts
Keyrxng: I've updated it to include one change made in #5 now checking for `pull_request` in the event. This helps restrict to only PRs and then further checks in the linked pr ensures the PR being closed is from the same repo as the current payload and belongs to the issue assignee

src/utils/issue.ts
Keyrxng: Must have been a typo as I know 100 is the max, surprised it didn't error but it likely defaults to max

src/utils/issue.ts
Keyrxng: So this should be a feature of the plugin logger and it should be configurable so as to ping the partner's TG and not Ubiquity TG. This will need to be well documented won't it so devs are aware fatal = tg shitstorm 😂

If a partner is running services through plugins and there is a fatal (not including the kernel because that should be ubiquity only) they should be notified in the same way?

Keyrxng:
Added ubiquity logger but the exported types are not the greatest. logger needs updated to have one entry file and expose everything that's needed in it, I will do this later today

src/utils/issue.ts
0x4007: Our partners shouldn't be expected to be doing maintenance on our product. Obviously that's ideal but that's unrealistic to expect.

README.md
```diff
@@ -1,32 +1,56 @@
...
+  command: "/(start|stop)"
+  example: "/start | /stop"
+  with:
+    disabledCommands: []
```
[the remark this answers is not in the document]
0x4007: What is this used for

src/adapters/index.ts
0x4007: > I've branched from the latest commit of this PR and threw something together quickly but I have a question
> 
> 
> 
> I'm assuming there would be a dedicated `ubiquibot-plugin-storage` repo the partner would create which would be a centralized place which plugins for that install could use?
> 

`.ubiquibot-config` repo 

> 
> The incoming payload has the repo in which the `/start` command was used so I can't read/write to every repo and create a `start-stop-storage.json` but I could from a dedicated repo easily. 
> 

Sorry but this doesn't make sense to me. Can you elaborate on why it's impossible to write to the event source repository?

> 
> The impl has pulled gh ids and wallets from our sb and has written to `Record<id, wallet>` then all sb logic removed and replaced with read/write/commit to repo but without the above being true, where exactly would each plugin store the data in the partner's org?
> 

Please never use acronyms in our technical discussions. 

1. It increases cognitive overhead 
2. It reduces clarity
3. It wastes time because now I need to ask you what is "impl" before I can provide you a useful response. 
4. Other contributors with less context are less likely to be able to contribute to the conversation. 

I would only accept the use of widely used acronyms that people understand without onboarding to our project. 

> 
> <hr/>
> 
> 
> 
> I think it's a good idea because any installed plugins by a partner should be trusted so each plugin having the ability to easily access the storage of other installed plugins makes things really easy for plugin DX and interoperability.
> 
> 
> 
> Even if the storage repo is private just knowing the storage structure of various plugins makes it easier to build on top of things. E.g. you could name the storage for this simply `wallets.json` and then other plugins need to confirm a registered user for that org checks `wallets.json`, further, `/wallet` could read/write to the same file separating logic but sharing storage 
> 
> 
> 
> <hr/>
> 
> 
> 
> I forget the term you used before but essentially the pre-packaged plugins we'd advise partners to install. You'd have the bare essentials like `/start|/stop`, `/wallet` etc so with that they'd each define a spot in the storage repo which could have rules like you mentioned so it can be modified or removed but it cannot be renamed and must adhere to a specific data structure, otherwise, go nuts

src/utils/issue.ts
Keyrxng: I agree ideal but unlikely however I meant in the sense that essentially we'd be a service provider like Cloudflare for example and if your services go down then you are notified as a service user, sometimes far later than you want to be.

How is that being addressed for partners if one of the services we provide hits a fatal for example. Manually case-by-case only if it's not fixed within a set time? Or should they also be included in this TG ping, would they have a dedicated channel we could ping maybe?

src/utils/get-linked-prs.ts
gentlementlegen: The main issue I had was that it was not reliable as `cross-referenced` just meant quoted by another PR that may or may not close it, so ended up evaluating comments that were not relevant. Still baffled that there is not easy way to query linked PRs through the API.
I ended up parsing the body of the linked pull-request itself, but if your approach is more reliable it'd be nice to use it inside the other plugins.

src/utils/issue.ts
gentlementlegen: `fatal` could mean incidents that we should resolve, but I think this would belong to the kernel directly since it is only a concern for us and we don't want to implement it inside every plugin.

package.json
gentlementlegen: So far I had an okay experience with `bun` except for testing. `yarn@4` is fine to me as long as we don't use `Zero Install` which implies committing the whole node modules list. The main concern with yarn > 1 (which also applies to bun actually) is compatibilty with tools. For example, Vercel would accept neither (only yarn classic, npm, pnpm), so would AWS etc. Github actions don't seem to mind, neither does Cloudflare.

src/handlers/shared/structured-metadata.ts
gentlementlegen: Isn't this solved, at least partially, by the updates you did to `ubiquibot-logger` @Keyrxng ?

Keyrxng: from a non-comment posting standpoint yes more or less this logic exists in `Logs.ts` and is returned under `metadata` in the `LogReturn` object from any Log call. It includes manipulating the stack etc but it doesn't include logic for formatting to hide content in a comment it's just a diff at the minute.

But you make a good point in that it could also just be handled by the logger which is expected to be used in all plugins

src/utils/get-linked-prs.ts
Keyrxng: Aye my heart was in my mouth the first time I saw it try to close this PR lol. From the QA I performed on test issues it does appear to be better yeah but only with restricting it further to match the payload repo etc

src/utils/issue.ts
Keyrxng: Do you think there should be two loggers, one that does not include fatal and is meant for plugins while the other can have Supabase and Telegram integrations that's more for the kernel?

rndquu: There are 3 entities here:
1. Kernel
2. Core ubiquity plugins
3. 3rd party plugins

Kernel and core ubiquity plugins should share the same logger instance. So if `logger.fatal()` is called in the kernel or core plugin then it's definitely an issue that we should resolve ASAP.

3rd party plugin developers should be responsible for their own plugins and maintain their own logging and monitoring solutions.

There is no need to add logger to the SDK thus allowing 3rd party plugins to use our own logging solution.

It makes sense to create a separate logger package (which is already [created](https://github.com/ubiquity/ubiquibot-logger) as far as I understand) and use it inside the kernel and core plugins.

Keyrxng: The logger package was just stripped down to work with 3rd party better than with the kernel and core plugins which is why I thought two loggers made sense

So 3rd party use their own and the template should use the basic `logger.warn() => console.warn()` setup it was originally and the logger refactored for Supabase/Telegram posting

gentlementlegen:
@Keyrxng CSpell workflow will crash because of yarn version not defined within the workflow
```
error This project's package.json defines "packageManager": "yarn@4.2.2". However the current global version of Yarn is 1.22.22.
```

wrangler.toml
```diff
@@ -0,0 +1,6 @@
+name = "ubiquibot-start-stop"
```
gentlementlegen, line 1:
```suggestion
name = "ubiquibot-command-start-stop"
```

src/types/plugin-input.ts
```diff
@@ -0,0 +1,35 @@
...
+  eventPayload: TU["payload"];
+  settings: StartStopSettings;
+  authToken: string;
+  ref: string;
+}
+
+const ONE_DAY = 24 * 60 * 60 * 1000;
+
+export const startStopSchema = T.Object({
+  isEnabled: T.Boolean(),
```
gentlementlegen, line 17:
```suggestion
  isEnabled: T.Boolean({ default: true} }),
```

I think it makes more sense to enable it by default.

gentlementlegen:
image

Just tested again an it seems there would be some formatting issues.

Second thing is a problem with the linked PRs as seen in [this test](https://github.com/Meniole/todel/issues/6#issuecomment-2212282816). This one should have been filtered out.

Otherwise the logic is working.

0x4007:
> image
> Just tested again an it seems there would be some formatting issues.

This has always existed. It's fine to ship like this until we can figure a solution later (I couldn't)

> Second thing is a problem with the linked PRs as seen in [this test](https://github.com/Meniole/todel/issues/6#issuecomment-2212282816). This one should have been filtered out.
> 
> Otherwise the logic is working.

gentlementlegen:
@0x4007 I mentioned it because I think it worsened, if you look at our current bot
image
so I thought it'd be easier to track the root of the issue. But it can be a separate issue, surely this is not urgent.

gentlementlegen:
@Keyrxng Let's have those fixed in a separate PR.

gentlementlegen:
@Keyrxng If you could address the remaining comments so we can merge this would be nice 😄

Keyrxng:
Latest QA

- https://github.com/ubq-testing/command-start-stop/issues/1#issuecomment-2222953485

> Not sure if this is better, but I remember seeing this render in a diff block so there's no need to dress it up if formatting isn't supported.

It was previously rendered in a diff block and formatting is not supported. I have listed them to support formatting with your suggestion but without the diff effects. Should it be rendered in a diff block?

- https://github.com/ubq-testing/command-start-stop/issues/2#issuecomment-2222964355

0x4007:
Match as close as possible to how it was and I can fix it later or something

Keyrxng:
@0x4007  https://github.com/ubq-testing/command-start-stop/issues/2#issuecomment-2223403247

src/types/context.ts
```diff
@@ -3,7 +3,7 @@ import { Octokit } from "@octokit/rest";
 import { StartStopSettings } from "./plugin-input";
 import { createAdapters } from "../adapters";
 import { Env } from "./env";
-import { Logs } from "@ubiquity-dao/ubiquibot-logger/.";
+import { Logs } from "@ubiquity-dao/ubiquibot-logger/";
```
gentlementlegen, line 6:
```suggestion
import { Logs } from "@ubiquity-dao/ubiquibot-logger";
```

The slash should not be required right?

gentlementlegen (approved):
Please fix my last comment and good to go for me.

src/types/context.ts
0x4007: Implies index.ts I presume

Keyrxng: > Implies index.ts I presume (which is not convention)

I don't understand what you mean here not convention

> The slash should not be required right?

No It's not. One single `index.ts` files collects the important exports and makes them available via `...logger(/ | /. | /index`)

Keyrxng: @0x4007 do you mean you'd rather have `/logger/logs` than just `/logger`?

Imo packages with many exports and distinct sections benefit more from have multiple entries (e.g. `permit-generation` (which needs fixed export-wise, haven't got to it yet and low priority)) 

But with only a few exports it's more convenient to have a single entry.

0x4007 (approved):

Keyrxng:
Would I merge this since everyone has approved or are we still waiting on @rndquu?

gentlementlegen:
@Keyrxng Should be good to go.

PR #7, opened 2024-07-11
Manifest file

gentlementlegen:
Depends on #1 
Relates to https://github.com/ubiquity/ubiquibot-kernel/pull/71


manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Start | Stop",
+  "description": "Assign or un-assign yourself from an issue.",
+  "commands": {
+    "start": {
+      "ubiquibot:example": "/start",
```
0x4007, line 6:
```suggestion
      "ubiquity:example": "/start",
```

I want to deprecate "ubiquibot" across all of our code and marketing materials in favor of new terminology. I think "ubiquity" is the safest and will always be relevant so that is my default here. 

However as a heads up I was thinking of referring to the system as "UbiquityOS" or "superintendent" in reference to the kernel.

0x4007 (changes requested):
Check the files view. There's some unused types.

gentlementlegen:
> Check the files view. There's some unused types.

I think these are false positives by Knip. `startStopSettingsValidator` is used [here](https://github.com/ubiquibot/command-start-stop/blob/c63579b364553781931f3c85515a009074c0c5e1/src/worker.ts#L26) and `PluginInputs` [there](https://github.com/ubiquibot/command-start-stop/blob/c63579b364553781931f3c85515a009074c0c5e1/src/plugin.ts#L8).

gentlementlegen:
@0x4007 I fixed workflows, everything will be green after this is merged to development.

PR #9, opened 2024-07-14
Max assignments

### LINKED ISSUE #8: Max Tasks
The old version has the ability to set limits. 

Based on the contributor level (collaborator, admin etc) we should be able to configure limits. 

Block start if they are assigned to more than X tasks. 

This is useful for new people who self assign a ton of tasks and of course can't handle all of them so just blocks progress.

jordan-ae:
Resolves #8

This functionality prevents new or less experienced contributors from overcommitting and blocking progress by self-assigning too many tasks. If the user is a core team member there's no max task limit.

QA: [ubq-test-jordan](https://github.com/ubq-test-jordan/command-start-stop/issues/19#issuecomment-2246149351)


jordan-ae:
@Keyrxng & @gentlementlegen Not sure which Issue to open this PR against. Should I open it against [this command-start-stop repo](https://github.com/ubq-testing/command-start-stop). The changes that I made are very few

gentlementlegen:
@jordan-ae Please rebase, I just merged https://github.com/ubiquibot/command-start-stop/pull/1 in.

jordan-ae:
Thanks @gentlementlegen all good now

Keyrxng:
I will review this today @jordan-ae

src/handlers/shared/check-org-member.ts
```diff
@@ -0,0 +1,9 @@
+import { Context } from "../../types";
+
+export async function isUserMember(context: Context, user: string) {
```
gentlementlegen, line 3: Could have a more explicit name, showing that it checks for the user's organization

src/handlers/shared/start.ts
```diff
@@ -44,7 +46,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
 
   // check for max and enforce max
 
-  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
+  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks && !isUserMember(context, assignee)) {
```
gentlementlegen, line 49: According to the spec, this should be configurable.

gentlementlegen (changes requested):

gentlementlegen:
It would be nice to have related unit tests regarding the configuration settings. Also, the max value should be configurable based on the user's association level (collaborator, contributor, member etc.).

Keyrxng:
+1 to the above comments.

I'd also keep in mind that custom roles can be defined within an org so it should be able to cover those too

src/handlers/shared/check-org-member.ts
jordan-ae: I'll replace with this name `get-user-role-in-org`. Any better suggestions please?

src/handlers/shared/start.ts
jordan-ae: about the configurability, what are the various user roles possible and what should be their maxConcurrentTasks limits. Please

jordan-ae:
> +1 to the above comments.
> 
> I'd also keep in mind that [custom roles](https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization) can be defined within an org so it should be able to cover those too

Hummm interesting do we have like standard user roles that Ubiquity uses, or we do I have to account for all the possible custom roles that can possibly be

Keyrxng:
> > +1 to the above comments.
> > I'd also keep in mind that [custom roles](https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization) can be defined within an org so it should be able to cover those too
> 
> Hummm interesting do we have like standard user roles that Ubiquity uses

Ubiquity tends to use the pre-defined roles. 

> do I have to account for all the possible custom roles that can possibly be

Yeah just I'd just pass in an iterable type and parse whatever it defines, you don't need to know the roles as you write the code if the config does the heavy lifting

src/plugin.ts
```diff
@@ -1,12 +1,19 @@
...
 import { Env, PluginInputs } from "./types";
-
 import { Octokit } from "@octokit/rest";
 import { createClient } from "@supabase/supabase-js";
 import { createAdapters } from "./adapters";
 import { Logs } from "@ubiquity-dao/ubiquibot-logger";
 
+export type UserRole = "admin" | "member" | "contributor";
+
+export const maxTask: Record<UserRole, number> = {
```
Keyrxng, line 11: This should be passed in via the config settings

src/handlers/shared/start.ts
```diff
@@ -1,17 +1,19 @@
...
-import { isUserMember } from "./check-org-member";
 import { checkTaskStale } from "./check-task-stale";
 import { generateAssignmentComment } from "./generate-assignment-comment";
+import { getUserRole } from "./get-user-role";
 import structuredMetadata from "./structured-metadata";
 import { assignTableComment } from "./table";
 
 export async function start(context: Context, issue: Context["payload"]["issue"], sender: Context["payload"]["sender"]) {
   const { logger, config } = context;
-  const { maxConcurrentTasks } = config.miscellaneous;
```
Keyrxng, line 11: here is where you'd extract the configurable `maxTasks` object from

src/handlers/shared/get-user-role.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  try {
+    const response = await context.octokit.orgs.getMembershipForUser({
+      org: orgLogin,
+      username: user,
+    });
+
+    const role = response.data.role;
+
+    if (role === "admin") {
```
Keyrxng, line 18: Why don't you just return the role?

Keyrxng:
@gentlementlegen re: https://github.com/ubiquibot/command-start-stop/pull/9#issuecomment-2229372569 

Are we sticking with Knip in CI when it's error prone like this?

If we are then are we just manually ignoring the problem files?

src/handlers/shared/get-user-role.ts
```diff
@@ -0,0 +1,28 @@
+import { Context } from "../../types";
+
+export async function getUserRole(context: Context, user: string): Promise<string> {
```
Keyrxng, line 3: You shouldn't need to cast the value later if you annotated the return type as `Promise<UserRole>` here

src/plugin.ts
```diff
@@ -1,12 +1,19 @@
 import { Context } from "./types/context";
 import { userStartStop } from "./handlers/user-start-stop";
 import { Env, PluginInputs } from "./types";
-
 import { Octokit } from "@octokit/rest";
 import { createClient } from "@supabase/supabase-js";
 import { createAdapters } from "./adapters";
 import { Logs } from "@ubiquity-dao/ubiquibot-logger";
 
+export type UserRole = "admin" | "member" | "contributor";
```
Keyrxng, line 9: This would probably be a dynamic union based on the keys of the passed in settings or a type which contains the pre-defined roles and can be extended

Keyrxng:
> do we have like standard user roles that Ubiquity uses

@jordan-ae one thing I had trouble with building plugins at first is the mental model. While Ubiquity is currently the primary user of the bot and it's plugins, we are not the intended demographic for either.

Plugins are built with the "Partner" and their needs in mind which is why configurability needs to be handled in such a way that it's either real easy (on/off or a specific value(s)) or we handle it elegantly under-the-hood.

0x4007 plans to have complete configurability right down to the `/<cmd>` string without affecting plugin execution. So when it's things that can be arbitrary like labels or roles you need to think "how can I handle all possibilities?" otherwise we impose standards onto partners which could make the plugin less attractive.

gentlementlegen:
> @gentlementlegen re: [#9 (comment)](https://github.com/ubiquibot/command-start-stop/pull/9#issuecomment-2229372569)
> 
> Are we sticking with Knip in CI when it's error prone like this? I know rndquu has mentioned it before also
> 
> If we are then are we just manually ignoring the erroneous problem files?

I think the only valid file to ignore is the entry point. I actually solve that by mentioning the entry point within the `package.json` so Knip doesn't complain and evaluates everything correctly.

jordan-ae:
I've been dreading sending in these changes because I'm unsure if what I've done makes sense. Essentially, I've modified the `maxConcurrentTask` configuration to accept an array of roles and limits. Then, I check the configuration against the user's role to determine the appropriate maximum number of tasks.

src/handlers/shared/get-user-role.ts
jordan-ae: at second glance it looks so dumb😅! I've modified it now

jordan-ae:
> > do we have like standard user roles that Ubiquity uses
> 
> @jordan-ae one thing I had trouble with building plugins at first is the mental model. While Ubiquity is currently the primary user of the bot and it's plugins, we are not the intended demographic for either.
> 
> Plugins are built with the "Partner" and their needs in mind which is why configurability needs to be handled in such a way that it's either real easy (on/off or a specific value(s)) or we handle it elegantly under-the-hood.
> 
> 0x4007 plans to have complete configurability right down to the `/<cmd>` string without affecting plugin execution. So when it's things that can be arbitrary like labels or roles you need to think "how can I handle all possibilities?" otherwise we impose standards onto partners which could make the plugin less attractive.

Thanks @Keyrxng this helped a lot

src/handlers/shared/start.ts
```diff
@@ -13,7 +12,8 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   const { taskStaleTimeoutDuration } = config.timers;
   const assignee = context.payload.issue.user.login;
   const userRole = (await getUserRole(context, assignee)) as UserRole;
-  const maxConcurrentTasks = maxTask[userRole];
+  const { maxConcurrentTasks } = config.miscellaneous;
+  const maxTask = maxConcurrentTasks.find(({ role }) => role.toLowerCase() === userRole)?.limit as number;
```
Keyrxng, line 16: How do we handle the case if no role is found? Can we provide some feedback to the user that they haven't configured this role? Do we then default to the lowest default role?

src/types/plugin-input.ts
```diff
@@ -13,6 +13,10 @@ export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU
 
 const ONE_DAY = 24 * 60 * 60 * 1000;
 
+const userRoleSchema = T.Union([T.Literal("Admin"), T.Literal("Member"), T.Literal("Contributor")]);
```
Keyrxng, line 16: I think this might prohibit role extension since this is a literal union of three items when it could potentially be something else

src/types/plugin-input.ts
```diff
@@ -24,10 +28,33 @@ export const startStopSchema = T.Object({
...
+        T.Object({
+          role: userRoleSchema,
+          limit: T.Integer(),
+        })
+      ),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {
+        maxConcurrentTasks: [
```
Keyrxng, line 41: These defaults are good and should kick in if the user does not inset any role info into their config.

I think I'd raise "member" to 10 as that's the core contributor/team members and they'd likely be on par with how many issues an admin can take too realistically but @gentlementlegen can confirm these figures.

src/types/plugin-input.ts
```diff
@@ -24,10 +28,33 @@ export const startStopSchema = T.Object({
   ),
   miscellaneous: T.Object(
     {
-      maxConcurrentTasks: T.Number(),
       startRequiresWallet: T.Boolean(),
+      maxConcurrentTasks: T.Array(
+        T.Object({
+          role: userRoleSchema,
```
Keyrxng, line 34: What happens if I add a role into my config like "newbie" or "gilded" and my org has it too, would this produce a runtime error because `userRoleSchema` only supports the three you've listed?

[this](https://docs.github.com/en/enterprise-cloud@latest/admin/overview/setting-up-a-trial-of-github-enterprise-cloud) trial would allow proper QA as it allows you to create custom roles as an enterprise otherwise unit tests would suffice

[This endpoint](https://docs.github.com/en/rest/orgs/organization-roles?apiVersion=2022-11-28#get-all-organization-roles-for-an-organization) returns the roles that an organization has and the response schema shows how varied it might be. Consider this, and extend the typebox schemas so that they are fully extendable

Keyrxng:
> what I've done makes sense

It does make sense yeah you have the right approach now. The code is starting to look acceptable, it would be good to see some QA of the changes in action as well as some unit tests covering the added configurables

src/handlers/shared/start.ts
jordan-ae: For now the default lowest role is contributor

Keyrxng: Yeah but keep in mind those only take affect if the user inputs no role definitions. If they define three custom roles and the role returned is not one of the three this should be handled via logging feedback and then finding the lowest limit in the config and applying that.

0x4007:
@Keyrxng  let's add support for custom roles in a future task if it's only for enterprise. It's not something we need or can easily test for as you mention.

Keyrxng:
@jordan-ae given the above comment what to do is:

- add default support for the other pre-defines such as `billing_manager`
- assume the roles will always be non-null and will only use the pre-defined roles
- if the role returned by the function call doesn't match any from the config then apply the lowest value limit from the config and log feedback

provide QA by either extending the unit tests or manually as I have done before with an alt account on test issues with your org and applying the various predefined roles to the alt, whichever is easier for you

package.json
```diff
@@ -84,6 +84,5 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  },
-  "packageManager": "yarn@4.2.2"
```
gentlementlegen, line 88: Any reason to remove the package manager for this pull-request?

src/types/plugin-input.ts
```diff
@@ -13,6 +13,10 @@ export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU
 
 const ONE_DAY = 24 * 60 * 60 * 1000;
 
+const userRoleSchema = T.String();
```
gentlementlegen, line 16: User role should be a `T.Literal` most likely

Keyrxng: @gentlementlegen are you suggesting that for this iteration @jordan-ae should define each pre-defined role as you say which was similar to how he had originally and any future custom role logic can refactor things later but for now be explicit and define each role?

gentlementlegen: Maybe the best choice would be to have a `T.Union` with `T.Literal` and `T.String`. The idea is to have a better auto-complete and less error prone type.

package.json
jordan-ae: Had an issue with package manager. Was removing the specified version will working and putting it back before pushing. It slipped my mind to put it back in this time. Apologies. Will Fix!

jordan-ae:
I have added the default GitHub roles and included T.String() with the default as Literals, ensuring that the action also accounts for custom types. For the default type, the implementation now selects the lowest limit value if there is no match with any of the defaults in the configuration

jordan-ae:
@Keyrxng concerning QA do I have to send like screenshots or like a loom of the alt account test?

src/handlers/shared/get-user-role.ts
```diff
@@ -2,19 +2,26 @@ import { Context } from "../../types";
...
 export async function getUserRole(context: Context, user: string): Promise<string> {
   const orgLogin = context.payload.organization?.login;
+  const { config } = context;
+  const { maxConcurrentTasks } = config.miscellaneous;
 
   if (!orgLogin) {
     throw new Error("Organization login not found in context payload.");
   }
 
-  try {
```
Keyrxng: Most if not all other octokit calls are wrapped with a try catch, let's keep this one wrapped too

src/handlers/shared/get-user-role.ts
```diff
@@ -2,19 +2,26 @@ import { Context } from "../../types";
...
+  });
 
-    return response.data.role;
-  } catch (error) {
-    return "contributor";
+  const matchingUser = maxConcurrentTasks.find(({ role }) => role.toLowerCase() === response.data.role);
+
+  if (matchingUser) {
+    //chech if the current user role matches any of those defined in the config
+    return matchingUser.role;
```
Keyrxng, line 21: I think we can return `matchingUser` and `minLimitTask` which gives us the limit too

src/handlers/shared/start.ts
```diff
@@ -13,7 +13,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   const assignee = context.payload.issue.user.login;
   const userRole = await getUserRole(context, assignee);
   const { maxConcurrentTasks } = config.miscellaneous;
-  const maxTask = maxConcurrentTasks.find(({ role }) => role.toLowerCase() === userRole)?.limit as number;
+  const maxTask = maxConcurrentTasks.find(({ role }) => role.toLowerCase() === userRole.toLowerCase())?.limit as number;
```
Keyrxng, line 16: this isn't needed as you are doing this inside `getUserRole()` already, you can take the `limit` from the function return data

src/handlers/shared/start.ts
```diff
@@ -48,7 +48,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
 
   // check for max and enforce max
 
-  if (assignedIssues.length - openedPullRequests.length >= maxTask) {
+  if (assignedIssues.length - openedPullRequests.length >= 3) {
```
Keyrxng, line 51: this should be something like `maxTasks.limit` or however you decide to name the `getUserRole()` return var

src/handlers/shared/get-user-role.ts
Keyrxng: I'm sure I received the exact same advice from someone early on in the team too, I have a clear memory of it but can't find the comment 😂

Keyrxng:
> @Keyrxng concerning QA do I have to send like screenshots or like a loom of the alt account test?

@jordan-ae typically we'd expect to see a live action demo of a plugin and the changes made via a link to one of your org issues showing it in action especially if unit tests aren't included but usually both are

For something like this manually, I'd assign my alt account roles and max the tasks and screenshot logs showing the role and it being declined because of x assigned tasks then I'd link the issue/comment for each role I was testing.

Automated you could leverage the mock db and make a setup function to create the amount of assigned tasks needed for each role, so long as they exist in the db assigned before you call `/start` then tests should pass. (not a how-to but a general approach as some http handlers may need updated)

Below is a QA link for an open PR of mine for reference

https://github.com/ubq-testing/command-start-stop/issues/5#issuecomment-2243867870

src/handlers/shared/start.ts
jordan-ae: Removed Thanks!

src/handlers/shared/get-user-role.ts
jordan-ae: Done Thanks!

jordan-ae:
So far I think everything code wise is good. I've added the QA too. The only thing I cant figure out is why all test concerning /start command keep failing no matter what I do.

Keyrxng:
> So far I think everything code wise is good. I've added the QA too. The only thing I cant figure out is why all test concerning /start command keep failing no matter what I do.

Look at this [run](https://github.com/ubiquibot/command-start-stop/actions/runs/10064886988/job/27822946199?pr=9) it shows why the tests are failing. You need to include a http handler that covers that endpoint so that we mock the response in tests.

gentlementlegen:
@jordan-ae any news on this task?

jordan-ae:
> @jordan-ae any news on this task?

@gentlementlegen I'm having a very hard time with test. Haven't written a lot of Jest test before so it's not very easy

Keyrxng:
> > @jordan-ae any news on this task?
> 
> @gentlementlegen I'm having a very hard time with test. Haven't written a lot of Jest test before so it's not very easy

@jordan-ae I'm unsure what exactly is holding you back whether it's still the same error or if you are further on from that point now.

I am happy to help if you can give more info. I have previously been in the same boat and what I did was went and took a look at other PRs in the same repo which might be making similar changes.

tests/__mocks__/handlers.ts
```diff
@@ -16,6 +16,20 @@ export const handlers = [
     }
     return HttpResponse.json(item);
   }),
+  //get member
+  http.get("https://api.github.com/orgs/:org/memberships/:username", ({ params: { username } }) => {
```
gentlementlegen, line 20: Maybe it'd be better to seed these values within the database mock

tests/main.test.ts
```diff
@@ -17,8 +17,9 @@ type Sender = Context["payload"]["sender"];
 
 const octokit = jest.requireActual("@octokit/rest");
 
-const url = process.env.SUPABASE_URL;
-const key = process.env.SUPABASE_KEY;
+const url = "https://sbxscxthunnhozvgtshx.supabase.co";
```
gentlementlegen, line 20: We should not use real endpoints. If it is to put dummy values, then set something that doesn't exists. If it is to actually reach a db, you have 2 options:

- mock Supabase library (efficient but tedious)
- intercept requests with `msw` (longer to do but easier to use on the long run from my experience)

src/types/plugin-input.ts
```diff
@@ -24,10 +26,33 @@ export const startStopSchema = T.Object({
...
+        T.Object({
+          role: userRoleSchema,
+          limit: T.Integer(),
+        })
+      ),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {
+        maxConcurrentTasks: [
```
gentlementlegen, line 39: This default is not proper since `maxConcurrentTask` should have its own default, not inside the parent object.

tests/main.test.ts
```diff
@@ -17,8 +17,9 @@ type Sender = Context["payload"]["sender"];
...
+const key =
+  "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6InNieHNjeHRodW5uaG96dmd0c2h4Iiwicm9sZSI6ImFub24iLCJpYXQiOjE3MTE0Nzg2MDcsImV4cCI6MjAyNzA1NDYwN30.-DstUzIspsRlLQPHR1da6nY5GBBs5I4Wd0jlK5zFzyU";
```
0x4007: Is this a real key? 👀

gentlementlegen: Looks like a public key, and that URL is not under our org. But we should avoid committing these anyway.

jordan-ae: I've gotten rid of it

tests/main.test.ts
```diff
@@ -195,51 +195,41 @@ describe("User start/stop", () => {
...
       if (error instanceof Error) {
         expect(error.message).toEqual("Too many assigned issues, you have reached your max limit of 3 issues.");
       }
     }
   });
+
+  test("should set maxLimits to 5 if the user is a member", async () => {
+
+    const issue = db.issue.findFirst({ where: { id: { equals: 7 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 1 } } }) as unknown as Sender;
```
Keyrxng, line 219: This doesn't feel right.

`sender === 1 === "ubiquity" === "admin"`

If sender is an admin and they are sending the `/start` command then the error below should match the config value which is `10`

tests/main.test.ts
```diff
@@ -382,6 +372,24 @@ async function setupTests() {
     owner: "ubiquity",
     repo: "test-repo",
   });
+
+  db.issue.create({
+    ...issueTemplate,
+    id: 7,
+    assignee: {
```
Keyrxng, line 379: You have created two new issues which are currently assigned to users who only exist here, I can't find mention of them anywhere else. If that is the case then how can they have reached their max assigned tasks?

If the issue is already assigned when someone calls `/start` on it (and you have loaded these new issues into your tests) then shouldn't it be handled with `This task is already assigned. Please choose another task.`?

tests/main.test.ts
```diff
@@ -195,51 +195,41 @@ describe("User start/stop", () => {
...
 
-    context.adapters = createAdapters(getSupabase(), context as unknown as Context);
+    context.adapters = createAdapters(getSupabase(), context);
 
     try {
-      await userStartStop(context as unknown as Context);
+      await userStartStop(context);
     } catch (error) {
       if (error instanceof Error) {
         expect(error.message).toEqual("Too many assigned issues, you have reached your max limit of 3 issues.");
```
Keyrxng, line 211: How can this error be a max limit of `3` when the sender here is an admin?

src/handlers/shared/get-user-role.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+
+    const matchingUser = maxConcurrentTasks.find(({ role }) => role.toLowerCase() === response.data.role);
+
+    if (matchingUser) {
+      //chech if the current user role matches any of those defined in the config
+      return matchingUser;
+    } else {
+      //return the role with the smallest task limit
+      return maxConcurrentTasks.reduce((minTask, currentTask) => (currentTask.limit < minTask.limit ? currentTask : minTask));
```
Keyrxng, line 26: We could move this outside of the try block so that we always return something from this function instead of throwing an error, which is the intent of this line anyway.

tests/__mocks__/handlers.ts
Keyrxng: I think @gentlementlegen means to update the user entities inside the `user-get.json` so they are loaded in the db and then pull the user from the db here and return only the role

Keyrxng (changes requested):
I think what's happening is that the try catch blocks are not hitting because `sender 1` is an admin and does not have more than 10 issues assigned as only 8 exist.

If I am mistaken, please use a const and store the error string from inside the catch block and then check your expect case at the bottom of the test body outside of the catch

tests/main.test.ts
jordan-ae: I made some changes regarding these this morning please can you check them out?

Keyrxng: As far as I can tell you have removed the new issues you created and have created two new users in the DB that have no assigned issues, so they'll pass your validation checks.

If there is a config limit of 10 for admins, then there would need to be at least ten issues right? Or we could lower the limit we test with and then have logs that we use `expect()` on and confirm the input/output/outcome specifically so there is no doubt

jordan-ae: I've made changes to match your recommendations. I've added issues to the users I created and tested the test fail when they should and pass when the should.

tests/main.test.ts
jordan-ae: Fixed!

src/handlers/shared/get-user-role.ts
jordan-ae: Makes sense!

tests/main.test.ts
```diff
@@ -208,12 +209,12 @@ describe("User start/stop", () => {
       await userStartStop(context);
     } catch (error) {
       if (error instanceof Error) {
-        expect(error.message).toEqual("Too many assigned issues, you have reached your max limit of 3 issues.");
+        expect(error.message).toEqual("Too many assigned issues, you have reached your max limit of 2 issues.");
```
gentlementlegen, line 251: Ideally this should be picked up from the configuration or the test so the string is dynamic maybe, so if we change it the tests doesn't fail? Forget about this if it's too much work.

Keyrxng: @gentlementlegen I think it's best that @jordan-ae also does as I was advised by whilefoo in my PR. My implementing of the tests in this way above was poor and should not be used.

@jordan-ae you don't have to change any other tests just the new ones that you are adding/refactoring make sure they use the below format.

```ts
const {config} = context
const adminLimit = config.maxTasks[0].limit

await expect(userStartStop(context)).rejects.toThrow(`Too many assigned issues, you have reached your max limit of ${adminLimit} issues.`);
```


[my pr for reference](https://github.com/ubiquibot/command-start-stop/pull/19/commits/5d487fb3d14cb369c9f47cde5fc3fde9715168a6#diff-12cd00a89af03f1fc6f74460b886289151edce062c60319b0c58fb8465558ed6)

Keyrxng:
While manually QA-ing I noticed issues with fetching assigned issues and fetching pull reviews. As to save time I have opened https://github.com/ubq-test-jordan/command-start-stop/pull/21 which resolves these issues and updates the tests accordingly

src/handlers/shared/get-user-role.ts
```diff
@@ -0,0 +1,25 @@
+import { Context } from "../../types";
+
+interface MatchingUserProps {
+  role: string;
+  limit: number;
+}
+
+export async function getUserRole(context: Context, user: string): Promise<MatchingUserProps> {
```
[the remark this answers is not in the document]
jordan-ae: I've changed the name to `get-user-task-limit-by-role.ts` . I think this is  as expressive as it could get. It specifies exactly what the function does and what it returns

src/handlers/shared/start.ts
```diff
@@ -37,21 +38,18 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
-  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
-    const log = logger.error("Too many assigned issues, you have reached your max limit", {
-      assignedIssues: assignedIssues.length,
-      openedPullRequests: openedPullRequests.length,
-      maxConcurrentTasks,
-    });
-    await addCommentToIssue(context, log?.logMessage.diff as string);
-    throw new Error(`Too many assigned issues, you have reached your max limit of ${maxConcurrentTasks} issues.`);
+  if (Math.abs(assignedIssues.length - openedPullRequests.length) >= maxTask.limit) {
+    await addCommentToIssue(context, "```diff\n! Too many assigned issues, you have reached your max limit.\n```");
```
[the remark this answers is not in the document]
jordan-ae: I noticed that the logger is being used mostly where the error message has variables in it. And in a couple of cases across the code the logger is not used where there a no variables to pass in the error message. So I think the trend is just being followed here.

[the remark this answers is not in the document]
Keyrxng: My mistake @gentlementlegen I think this may have been me, still forming the habit

@jordan-ae variables don't play into it, the logger takes the log and creates 4 versions of it, one of them is `diff` which is how the bot comments most of the time.

```ts
await addCommentToIssue(context, log?.logMessage.diff as string);
```

is the correct pattern

src/handlers/shared/get-user-task-limit-by-role.ts
```diff
@@ -0,0 +1,25 @@
+import { Context } from "../../types";
+
+interface MatchingUserProps {
+  role: string;
+  limit: number;
+}
+
+export async function getUserRole(context: Context, user: string): Promise<MatchingUserProps> {
```
gentlementlegen, line 8: This is called `getUserRole` but it returns a variable named `smallestTask`, I am a bit confused

src/handlers/shared/start.ts
```diff
@@ -37,21 +38,19 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   // check max assigned issues
 
   const openedPullRequests = await getAvailableOpenedPullRequests(context, sender.login);
-  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed: ${JSON.stringify(openedPullRequests)}`);
+  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed`, {
+    openedPullRequests: openedPullRequests.map((i) => i.html_url),
```
gentlementlegen, line 42:
```suggestion
    openedPullRequests: openedPullRequests.map((pr) => pr.html_url),
```

src/utils/issue.ts
```diff
@@ -127,37 +133,28 @@ export async function addAssignees(context: Context, issueNo: number, assignees:
...
 
-export async function getAllPullRequestReviews(context: Context, pullNumber: number, format: "raw" | "html" | "text" | "full" = "raw") {
-  const payload = context.payload;
-
-  const owner = payload.repository.owner.login;
-  const repo = payload.repository.name;
-
+export async function getAllPullRequestReviews(context: Context, pullNumber: number, owner: string, repo: string) {
   try {
-    return (await context.octokit.paginate(context.octokit.rest.pulls.listReviews, {
```
gentlementlegen, line 153: Should probably using `rest`

src/utils/issue.ts
```diff
@@ -174,7 +171,9 @@ export async function getAvailableOpenedPullRequests(context: Context, username:
 
   for (let i = 0; i < openedPullRequests.length; i++) {
     const openedPullRequest = openedPullRequests[i];
-    const reviews = await getAllPullRequestReviews(context, openedPullRequest.number);
+    const owner = openedPullRequest.html_url.split("/")[3];
+    const repo = openedPullRequest.html_url.split("/")[4];
```
gentlementlegen, line 175: Can be done in one instruction. What happens if the url is not from an organization?

src/handlers/shared/start.ts
```diff
@@ -37,21 +38,19 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
 
   // check for max and enforce max
 
-  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
-    const log = logger.error("Too many assigned issues, you have reached your max limit", {
-      assignedIssues: assignedIssues.length,
-      openedPullRequests: openedPullRequests.length,
-      maxConcurrentTasks,
-    });
+  if (Math.abs(assignedIssues.length - openedPullRequests.length) >= maxTask.limit) {
```
gentlementlegen, line 50: Wouldn't this mean that if you have no assigned issues but plenty of open pull-request, you would get limited even if you actually are assigned to nothing?

src/utils/issue.ts
```diff
@@ -1,26 +1,32 @@
...
+        q: `org:${payload.repository.owner.login} assignee:${username} is:open`,
         per_page: 100,
-      },
-      ({ data: issues }) => issues.filter((issue: Issue) => !issue.pull_request && issue.assignee && issue.assignee.login === username)
-    );
+        order: "desc",
+        sort: "created",
+      })
+      .then((issues) =>
+        issues.filter((issue) => {
```
gentlementlegen, line 22: Is the filter needed since the search specifies `is:open`? Might wanna add `is:pr` as well to the query.

src/handlers/shared/start.ts
Keyrxng: You raise a valid concern, but isn’t this approach actually preferable? Without this check, someone could bypass the task limit simply by opening multiple PRs. Essentially, this would make the limit meaningless.

Consider this scenario:

`maxLimit = 5`
`assignedIssues = 0`
`openPRs = 6`

The calculation would be `0 - 6 == -6` > `-6 + 5 == -1` > `-1 >= 5`. And for every new task, a new PR which means unlimited tasks.

So, with each additional PR, the max limit would increase, defeating the purpose of having a limit at all. This check ensures that the task limit is enforced properly, preventing such loopholes.

I could imagine a situation where a user starts a task, then opens a PR with some "filler" commits just to hold their place, or for some other strategic reason. This check helps prevent that kind of workaround.

gentlementlegen: I also agree with you. I guess any pull-request should have a related issue specification, I was thinking more when we open a bunch of PRs that actually are quick fixes. But I guess it is better to secure it that way.

src/utils/issue.ts
Keyrxng: > Is the filter needed since the search specifies is:open

no I guess not

> Might wanna add is:pr as well to the query.

This function is looking for issues not PRs so if anything `is:issue` should be added no, can we do `is:issue` yeah?

src/handlers/shared/start.ts
Keyrxng: You mean from a core team perspective I figure? In this case we could 

- bypass the check all together
- have org member checks configurable
- factor it into the config settings when defining `MEMBER` and have it be like 20-odds which seems reasonable
- ...

jordan-ae: With how the review process looks like here, that will make things really rough.

In a hypothetical scenario where a contributor has a limit of 3 task per time, successfully gets the tasks done (including test and everything related ) and the review process starts. From how the speed of the whole review process looks like here, it could take months to close PR's. It can get frustrating very fast. 

In the V1 bot, I thought this was a feature intentionally introduced to allow for continuation of work during the review process. Didn't know it was just a loophole we forgot to cover 😅

Keyrxng: In my experience, contributor's within the first 3-5 months likely cannot efficiently manage more than two or three concurrent tasks anyway. After this code review moves a lot faster and it is no longer such an issue for them, at least was the case for me and all I've seen this I started

> In the V1 bot, I thought this was a feature intentionally introduced to allow for continuation of work during the review process. Didn't know it was just a loophole we forgot to cover 

It was an intentional feature but it's being improved vis-a-vis `automated-merging`, review incentives etc. I'd personally rather see a contributor take a PR from 0 - 100 before moving on to the next task.

src/utils/issue.ts
gentlementlegen: I think it is redundant and error prone if we ever change the query later, since the same check is ran twice. Hopefully the API should be reliable.

I missed the `!` you are right, `is:issue` would be the right call.

jordan-ae:
@gentlementlegen  & @Keyrxng anything else I should fix up? Thanks

src/utils/issue.ts
```diff
@@ -1,26 +1,32 @@
...
   try {
-    return await context.octokit.paginate(
-      context.octokit.issues.listForRepo,
-      {
-        owner: payload.repository.owner.login,
-        repo: payload.repository.name,
-        state: ISSUE_TYPE.OPEN,
+    return await context.octokit
+      .paginate(context.octokit.search.issuesAndPullRequests, {
+        q: `org:${payload.repository.owner.login} assignee:${username} is:open`,
```
Keyrxng, line 16: `is:issue` in this query for issues as we were asked by gentlementlegen

src/utils/issue.ts
```diff
@@ -127,37 +133,28 @@ export async function addAssignees(context: Context, issueNo: number, assignees:
...
+export async function getAllPullRequests(context: Context, state: "open" | "closed" | "all" = "open", username: string) {
+  const { payload } = context;
 
   try {
-    return (await context.octokit.paginate(context.octokit.rest.pulls.list, {
-      owner: payload.repository.owner.login,
-      repo: payload.repository.name,
-      state,
+    return (await context.octokit.paginate(context.octokit.search.issuesAndPullRequests, {
+      q: `org:${payload.repository.owner.login} author:${username} state:${state}`,
```
Keyrxng, line 141: `is:pr` here too

src/utils/issue.ts
jordan-ae: When I add `is:issue` then there's no need to `!issue.pull_request &&` right? Since the query already filters for issues only.

gentlementlegen:
I think the `README` does not reflect the changes that were made to the configuration.

src/types/plugin-input.ts
gentlementlegen: @jordan-ae this has not been taken care of. If the configuration looks like
```yml
        miscellaneous:
          # maxConcurrentTasks: []
          startRequiresWallet: true
```
It crashes with the message
```
        "Error: Unable to decode value as it does not match the expected schema"
```
because no default value is set. Also, it seems to crash with an empty array.

jordan-ae:
@gentlementlegen I went with the option of adding a Decode to check for duplicate values. I didn't go with the option of using keys instead because it will result in refactoring all the work we have done so far from the configs to the test. Overall the Decode solution is as effective in my opinion and doesn't necessitate so much refactoring. Hope I have butchered the implementation though.

src/handlers/shared/start.ts
0x4007: I didn't scrutinize the code because I'm mobile but I can tell a little story about v1

---

We had this problem where contributors would self assign a ton of tasks and then block progress. So we made a limit of how many open tasks can be actively assigned. 

Then we realized that our review team was actually causing the delays, so we decided that after a 24 hour grace period of pending review, we would lift the limit by one. 

However if a reviewer requested changes on any pull, then the limit is reduced again, which forces the contributor to focus on the open pulls instead of starting new tasks. 

At a high level, we realized that delays can happen on both the contributor and reviewer side. This system seemed to work quite well in old v1, and we should reimplement.

Keyrxng:
> @gentlementlegen I went with the option of adding a Decode to check for duplicate values. I didn't go with the option of using keys instead because it will result in refactoring all the work we have done so far from the configs to the test. Overall the Decode solution is as effective in my opinion and doesn't necessitate so much refactoring. Hope I have butchered the implementation though.

@jordan-ae I believe @gentlementlegen was referring to using `T.Transform(SCHEMA).Decode((val) => {}).Encode((val) => {}`. Examples can be found within the Typebox docs and throughout various plugins but here is a [recent](https://github.com/ubiquibot/crypto-faucet-plugin/blob/574d9129f6d016ad1641562214d55071fc94589a/src/types/plugin-inputs.ts#L39) one I wrote. The approach you used is not typically how we'd decode. 

Further, I believe you could of just added the `new Set(taskRoles)` into `getUserRoleAndTaskLimit()` and achieve the same effect that your decode function has.

@gentlementlegen am I correct in the above? Also how should we handle duplicates if they exist? Check their values and default to the lowest and post a comment that's what we did and which items are duplicated? I feel throwing an error is not great and we could handle it more elegantly

gentlementlegen:
@Keyrxng yes you are correct. I would have used `T.Record` to avoid needed to check for duplicates in a list and avoid the need to use `Decode` while ensuring every item is unique. This would require not so much refractor except changing the list loop by `Object.entries` to loop on the key / values.

jordan-ae:
Tbh I'm not sure if what I've done here is right. But I've gone with your recommendations and used `T.Record` and refactored to loop by `Object.entries`. Hope I haven't blown this too. Fingers crossed

src/types/plugin-input.ts
```diff
@@ -27,37 +26,25 @@ export const startStopSchema = T.Object({
   miscellaneous: T.Object(
     {
       startRequiresWallet: T.Boolean(),
-      maxConcurrentTasks: T.Array(
+      maxConcurrentTasks: T.Record(
```
Keyrxng, line 29: If using a Record here the string is our `userRole` and it's value could be the limit instead of `{limit: 0}`. Finding the limit could also be direct access instead of using `find()`

jordan-ae: Makes sense done. Thanks

src/handlers/shared/get-user-task-limit-and-role.ts
```diff
@@ -21,11 +21,10 @@ export async function getUserRoleAndTaskLimit(context: Context, user: string): P
       username: user,
     });
 
-    const matchedTask = Object.entries(maxConcurrentTasks).find(
-      ([role]) => role.toLowerCase() === response.data.role.toLowerCase()
-    );
+    const role = response.data.role.toLowerCase();
+    const limit = maxConcurrentTasks[role.charAt(0).toUpperCase() + role.slice(1)];
```
Keyrxng, line 25: It seems you have lowercased `role` and then capitalize it to index `maxConcurrentTasks` in the this line.

We don't know how a partner would define their roles so it makes sense to normalize the octokit returned role and the config roles. Preferably using the `T.Tranform().Decode().Encode()` method mentioned previously and like you have for the octokit response as-is

jordan-ae: @Keyrxng sorry I'm abit confused on this one ?

Keyrxng:
```ts
const role = response.data.role.toLowerCase();
const limit = maxConcurrentTasks[role.charAt(0).toUpperCase() + role.slice(1)]
```

You make `role` lowercase and then you capitalize the first letter and use it to index `maxConcurrentTasks` which might not match how the partner has defined their roles in the config as they may be lowercase

I'd suggest using the method we discussed [here](https://github.com/ubiquibot/command-start-stop/pull/9#issuecomment-2290370173) to handle decoding or whichever method you think is right

jordan-ae: I realized I wasn't supposed to do all that gymnastics. I was just supposed to use the octokit returned role and the config role as is.

Keyrxng: You have defined defaults as being capitalized but how do we guarantee that the role defined in the config by partners is uppercase or lowercase and whatever the octokit response is going to be?

> We don't know how a partner would define their roles so it makes sense to normalize the octokit returned role and the config roles.

jordan-ae: You're correct that the code as written doesn't enforce consistent casing between the roles defined in the configuration and those returned by octokit. 

To ensure reliable matching between the roles, regardless of how they're cased in the configuration or the API response, we can normalize both the configuration keys and the API response role to a consistent format (e.g., all lowercase) before performing the lookup. Will that work?

README.md
```diff
@@ -40,8 +40,11 @@ To configure your Ubiquibot to run this plugin, add the following to the `.ubiqu
       reviewDelayTolerance: 86000
       taskStaleTimeoutDuration: 2580000
     miscellaneous:
-      maxConcurrentTasks: 3
-      startRequiresWallet: true # default is true
+      maxConcurrentTasks: # Default concurrent task limits per role.
+        Admin: 10
```
gentlementlegen, line 44: Should be lowercase

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
...
     {
-      maxConcurrentTasks: T.Number(),
       startRequiresWallet: T.Boolean(),
+      maxConcurrentTasks: T.Record(T.String(), T.Integer()),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {
+        maxConcurrentTasks: {
+          Admin: 20,
```
gentlementlegen, line 34: should be lower case as well

jordan-ae: Done

README.md
jordan-ae: Done

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
   ),
   miscellaneous: T.Object(
     {
-      maxConcurrentTasks: T.Number(),
       startRequiresWallet: T.Boolean(),
+      maxConcurrentTasks: T.Record(T.String(), T.Integer()),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {
```
gentlementlegen, line 32:
```suggestion
      default: {}
```

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
   ),
   miscellaneous: T.Object(
     {
-      maxConcurrentTasks: T.Number(),
       startRequiresWallet: T.Boolean(),
```
gentlementlegen, line 28:
```suggestion
      startRequiresWallet: T.Boolean({ default: true }}),
```

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
...
+      maxConcurrentTasks: T.Record(T.String(), T.Integer()),
```
gentlementlegen, line 29:
```suggestion
      maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
```

gentlementlegen:
@jordan-ae Please resolve the conflicts and the comments so this can be merged in.

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
...
   miscellaneous: T.Object(
     {
-      maxConcurrentTasks: T.Number(),
-      startRequiresWallet: T.Boolean(),
+      startRequiresWallet: T.Boolean({ default: true }}),
+      maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {}
```
gentlementlegen, line 32: This very likely doesn't compile.

src/types/plugin-input.ts
```diff
@@ -23,12 +25,21 @@ export const startStopSchema = T.Object({
...
+        },
+        startRequiresWallet: true,
+      },
+    }
   ),
 });
 
-export type StartStopSettings = StaticDecode<typeof startStopSchema>;
-export const startStopSettingsValidator = new StandardValidator(startStopSchema);
+export type StartStopSettings = Static<typeof startStopSchema>;
```
Keyrxng, line 44: What is the reason behind using `Static` vs the original `StaticDecode`?

src/types/plugin-input.ts
```diff
@@ -1,5 +1,5 @@
 import { SupportedEvents, SupportedEventsU } from "./context";
-import { StaticDecode, Type as T } from "@sinclair/typebox";
+import { Static, StaticDecode, Type as T } from "@sinclair/typebox";
```
Keyrxng, line 2: Is this `Static` import needed?

src/types/plugin-input.ts
```diff
@@ -13,6 +13,8 @@ export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU
 
 const ONE_DAY = 24 * 60 * 60 * 1000;
 
+const userRoleSchema = T.Union([T.Literal("Admin"), T.Literal("Member"), T.Literal("Contributor"), T.String()]);
```
Keyrxng, line 16: is this used anywhere? The literals differ from the typebox defaults so if it's used sync them up otherwise remove it

src/types/plugin-input.ts
```diff
@@ -23,12 +25,20 @@ export const startStopSchema = T.Object({
...
     {
-      maxConcurrentTasks: T.Number(),
-      startRequiresWallet: T.Boolean(),
+      startRequiresWallet: T.Boolean({ default: true }),
+      maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
     },
-    { default: { maxConcurrentTasks: 3, startRequiresWallet: true } }
+    {
+      default: {},
+        maxConcurrentTasks: {
```
gentlementlegen, line 33: This should be removed, this is not event part of the arguments of `Type.Object` proper options.

src/types/plugin-input.ts
```diff
@@ -23,12 +25,20 @@ export const startStopSchema = T.Object({
   ),
   miscellaneous: T.Object(
     {
-      maxConcurrentTasks: T.Number(),
-      startRequiresWallet: T.Boolean(),
+      startRequiresWallet: T.Boolean({ default: true }),
+      maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
```
gentlementlegen, line 29: Most likely we can make sure that the Integer value is greater than `0` because it wouldn't make sense.

jordan-ae:
@gentlementlegen  & @Keyrxng, sorry for the mess and loose ends in the last few commits. I'm trying to rebase all the divergences between this branch and the main branch, but I don't have full context on several commits, which are largely different from what my branch has.

Do you guys know any way to fix the conflicts more easily? Usually, I'd create a new branch from main and redo my changes there, then open a new PR, but I don't think that's an option in this case.

gentlementlegen:
@jordan-ae How complicated is the merge? When it gets too messy what I do is keep a diff of what I changed, accept all the incoming merges, and redo again my changes manually (if they are not too important). Otherwise it is too much headache.

jordan-ae:
Got the merge conflicts fixed. I kinda screwed the test up in the process. So I have to get those cleaned up.

src/plugin.ts
```diff
@@ -42,5 +42,8 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
 }
 
 function sanitizeMetadata(obj: LogReturn["metadata"]): string {
-  return JSON.stringify(obj, null, 2).replace(/</g, "&lt;").replace(/>/g, "&gt;").replace(/--/g, "&#45;&#45;");
+  return JSON.stringify(obj, null, 2)
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/--/g, "&#45;&#45;")
```
gentlementlegen, line 48: Why specifically two dashes are escaped, but not single ones?

src/types/plugin-input.ts
```diff
@@ -15,13 +15,13 @@ export const startStopSchema = T.Object(
...
     reviewDelayTolerance: T.String({ default: "1 Day" }),
     taskStaleTimeoutDuration: T.String({ default: "30 Days" }),
-    maxConcurrentTasks: T.Number({ default: 3 }),
     startRequiresWallet: T.Boolean({ default: true }),
+    maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
   },
   {
-    default: {},
-  }
+    default: {}
```
gentlementlegen, line 22: Is this Prettier correct?

src/plugin.ts
Keyrxng: One hyphen is grammatically correct while two is usually code-like isn't? I also removed the other escapes because it made the comment unreadable

Keyrxng:
> Got the merge conflicts fixed. I kinda screwed the test up in the process. So I have to get those cleaned up.

@jordan-ae what I tend to do after something like this or if there has been lots of changes/time since the last QA is, I reproduce live QA testing that old features are not broken and the new ones work as expected via test issues. 

I'd appreciate if you could do that too, showing that `/start` and `/stop` work as expected and your new `maxTask` in action. It takes a little longer but it's always quicker than having to double back to fix things and then produce the QA anyway.

Looking at the jest run it looks like tests are failing because the endpoint it's trying to hit is missing the repo name `/repos/ubiquity//pulls/1/reviews`. Log the DB entity and confirm the url is correct if it's not fix where the test issue is created otherwise there may be an issue in splitting the url. Ping me on tg if you need help

jordan-ae:
Screenshot from 2024-09-02 11-24-21

I'm trying to QA again and for some reason I keep getting a 400 error

Keyrxng:
> Screenshot from 2024-09-02 11-24-21
> 
> I'm trying to QA again and for some reason I keep getting a 400 error. Do any of you guys please have any ideas on why this might be?

Unsure based on just that image. The kernel would have logged something if it was related to fetching the manifest which is the only thing I can think of if the kernel is processing an event from github and passing it to the worker. I can only suggest adding logs throughout in order to find where the fault is

jordan-ae:
@Keyrxng I got it fixed! The `envConfigValidator` was expecting `APP_ID` as an env var. This was added in one of the merged commits and the `dev.vars.example` didn't specify that change so it kept falling at the env check

jordan-ae:
QA: To ensure [start/stop](https://github.com/ubq-test-jordan/command-start-stop/issues/25#issuecomment-2324978309) commands are working as expected


QA: To ensure [max-limits](https://github.com/ubq-test-jordan/command-start-stop/issues/33#issuecomment-2325016941) is respected

jordan-ae:
@Keyrxng  Been trying to trash these two failing test since afternoon. Still working on it

Keyrxng:
> @Keyrxng Been trying to trash these two failing test since afternoon. Still working on it

ping me on tg if you need a hand

src/handlers/shared/start.ts
```diff
@@ -33,6 +36,21 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     logger.error("Error while getting commit hash", { error: e as Error });
   }
 
+  // check max assigned issues
+  const openedPullRequests = await getAvailableOpenedPullRequests(context, sender.login);
```
Keyrxng, line 40: these checks can be removed since we process checks for each user now in `handleTaskLimitChecks()`

src/handlers/shared/start.ts
```diff
@@ -33,6 +36,21 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
+  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed`, {
+    openedPullRequests: openedPullRequests.map((pr) => pr.html_url),
+  });
+
+  const assignedIssues = await getAssignedIssues(context, sender.login);
```
Keyrxng, line 45: same here

src/handlers/shared/start.ts
```diff
@@ -33,6 +36,21 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
+  logger.info("Max issues allowed is", { limit: maxTask.limit, assigned: assignedIssues.length });
+
+  // check for max and enforce max
+
+  if (Math.abs(assignedIssues.length - openedPullRequests.length) >= maxTask.limit) {
+    throw new Error(`Too many assigned issues, you have reached your max limit of ${maxTask.limit} issues.`);
```
Keyrxng, line 51: same here

Keyrxng (changes requested):
Tests still failing too it appears. Managing okay jordan?

jordan-ae:
> Tests still failing too it appears. Managing okay jordan?

@Keyrxng I really wanted to clear this one off without buggin' you. But I'm stuck on the two failing test I have not been able to get through. Sure I'm missing something really basic

Keyrxng:
> without buggin' you.

It's a talent in and of itself to know when to ask for help but in all cases I'll never feel like you are bugging me dude, I'd rather you popped up than pulled your hair out for days.

src/handlers/shared/start.ts
```diff
@@ -36,18 +36,12 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
-  });
-
   const assignedIssues = await getAssignedIssues(context, sender.login);
   logger.info("Max issues allowed is", { limit: maxTask.limit, assigned: assignedIssues.length });
 
   // check for max and enforce max
 
-  if (Math.abs(assignedIssues.length - openedPullRequests.length) >= maxTask.limit) {
+  if (assignedIssues.length  >= maxTask.limit) {
     throw new Error(`Too many assigned issues, you have reached your max limit of ${maxTask.limit} issues.`);
```
Keyrxng, line 45: all of these checks should have been removed. Look at the logic inside the `handleTaskLimitChecks` function which we fire for each user

src/handlers/shared/start.ts
```diff
@@ -36,18 +36,12 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     logger.error("Error while getting commit hash", { error: e as Error });
   }
 
-  // check max assigned issues
-  const openedPullRequests = await getAvailableOpenedPullRequests(context, sender.login);
-  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed`, {
-    openedPullRequests: openedPullRequests.map((pr) => pr.html_url),
-  });
-
   const assignedIssues = await getAssignedIssues(context, sender.login);
```
Keyrxng, line 39: I'm pretty sure this can be removed as well as this is part of the checks function too

src/handlers/shared/start.ts
jordan-ae: You are right the `handleTaskLimitChecks` already covers all that. I've tested and it still works as expected.

jordan-ae:
> I'd rather you popped up than pulled your hair out for days.

My bad I'll ask for help earlier next time

Keyrxng:
> > I'd rather you popped up than pulled your hair out for days.
> 
> My bad I'll ask for help earlier next time

Shoot me a msg on TG whenever, happy to help.

> I've tested and it still works as expected.

All tests passing locally? CI still seems to be failing still for some reason

Keyrxng:
@jordan-ae can you push this forward

jordan-ae:
Screenshot from 2024-09-05 01-05-50

src/handlers/shared/start.ts
```diff
@@ -1,15 +1,18 @@
-import { Context, ISSUE_TYPE, Label } from "../../types";
+
```
gentlementlegen, line 1:
```suggestion
```

README.md
gentlementlegen, line 1: The example configuration seems to be improperly indented.

src/types/plugin-input.ts
gentlementlegen:
```suggestion
    default: {},
```

src/types/plugin-input.ts
```diff
@@ -15,14 +15,14 @@ export const startStopSchema = T.Object(
...
-    maxConcurrentTasks: T.Number({ default: 3 }),
     startRequiresWallet: T.Boolean({ default: true }),
+    maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
     emptyWalletText: T.String({ default: "Please set your wallet address with the /wallet command first and try again." }),
   },
   {
-    default: {},
-  }
+    default: {}
+  },
```
gentlementlegen, line 24:
```suggestion
  }
```

README.md
jordan-ae: Fixed

gentlementlegen:
I tested this and it seems to double post when you get above your task limit: https://github.com/Meniole/command-start-stop/issues/14

jordan-ae:
Got it fixed: [Fix Duplicates](https://github.com/ubq-test-jordan/command-start-stop/issues/36#issuecomment-2331224337)

src/handlers/shared/start.ts
```diff
@@ -141,12 +143,22 @@ async function handleTaskLimitChecks(username: string, context: Context, maxConc
...
+    
+    const log = logMessage 
+      ? logger.error(logMessage, {
+          assignedIssues: assignedIssues.length,
+          openedPullRequests: openedPullRequests.length,
+          maxConcurrentTasks,
+        })
+      : null;
+  
+    if (log && log.logMessage.diff) {
```
gentlementlegen, line 158: I think the code would be clearer if the whole logic was within a `if` block checking the `user === sender` to avoid the two ternaries above.

gentlementlegen (changes requested):

src/handlers/shared/start.ts
jordan-ae: Done

src/handlers/shared/start.ts
```diff
@@ -57,7 +59,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   const toAssign = [];
   // check max assigned issues
   for (const user of teammates) {
-    if (await handleTaskLimitChecks(user, context, maxConcurrentTasks, logger, sender.login)) {
+    if (await handleTaskLimitChecks(user, context, maxTask.limit, logger, sender.login)) {
```
Keyrxng, line 62: You have passed in the task limit for the `sender` for each call to this function essentially applying the same limit to all assignees in a team scenario as opposed to moving the call to `getUserRoleAndTaskLimit` into the checks function itself

src/handlers/shared/start.ts
```diff
@@ -141,12 +143,19 @@ async function handleTaskLimitChecks(username: string, context: Context, maxConc
...
-    await addCommentToIssue(context, log?.logMessage.diff as string);
+    if (username !== sender) {
+      const logMessage = `${username} has reached their max task limit`;
+      const log = logger.error(logMessage, {
+        assignedIssues: assignedIssues.length,
+        openedPullRequests: openedPullRequests.length,
+        maxConcurrentTasks,
+      });
+    
+      if (log?.logMessage?.diff) {
```
Keyrxng, line 154: We shouldn't need to optionally access the `LogReturn` object anymore, I double checked your logger version and it's up to date so this shouldn't be needed

Keyrxng (changes requested):

Keyrxng (changes requested):

Keyrxng:
Removed items from my review because #35 addresses the double comment posting (due to the catch-all), logReturn etc. 

Not sure why the error has changed in this PR for when the `APP_ID` is not set vs `development` but I'll fix correct anything in #35

src/utils/issue.ts
gentlementlegen, line 1: `Issue` import is not used after your changes, please remove it.

gentlementlegen:
@jordan-ae You should enable `Prettier` on save because many files are not properly formatted. Run `yarn format:prettier` to fix them all at once.

gentlementlegen (approved):

jordan-ae:
> Works for me, here is my QA: https://github.com/Meniole/command-start-stop/issues/20

Yaaayyyyy 🎉

README.md
```diff
@@ -38,7 +38,10 @@ To configure your Ubiquibot to run this plugin, add the following to the `.ubiqu
   with:
     reviewDelayTolerance: "3 Days"
     taskStaleTimeoutDuration: "30 Days"
-    maxConcurrentTasks: 3
+    maxConcurrentTasks: # Default concurrent task limits per role.
+      admin: 10
```
0x4007, line 42: I feel like admins shouldn't have limits but to be honest I feel this feature has no teeth because we can just use the GitHub UI. 

It would be more interesting if it unassigns when we self assign and we hit the limit.

0x4007:
According to the original conversation on the issue, admins should not have limits. Please remove from the default.

jordan-ae:
@0x4007 i tried using infinity for admins in the yaml config. But it didn't seem to be too reliable. And to the best of my knowledge that's the easiest way to define unlimited task to admins. Other than that I think it will just make sense and be much easier to give admins a larger  task limit. Like 100 maybe? To represent infinity

jordan-ae:
Okay got it I've set admin to Infinity by default. So admins will no longer have any limits

README.md
gentlementlegen: @0x4007 we can change UI assignments to follow the same logic as `/start` so user would get unassigned if they are doing too many tasks, even when forcing it through the UI.

src/types/plugin-input.ts
```diff
@@ -16,7 +16,7 @@ export const startStopSchema = T.Object(
     reviewDelayTolerance: T.String({ default: "1 Day" }),
     taskStaleTimeoutDuration: T.String({ default: "30 Days" }),
     startRequiresWallet: T.Boolean({ default: true }),
-    maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: 20, member: 10, contributor: 2 } }),
+    maxConcurrentTasks: T.Record(T.String(), T.Integer(), { default: { admin: Infinity, member: 10, contributor: 2 } }),
```
gentlementlegen, line 19: Could you not totally remove it?

0x4007 (approved):
I'm assuming that Infinity works

PR #10, opened 2024-07-15
fix: close only assignee PR

Keyrxng:
Resolves https://github.com/ubiquibot/command-start-stop/issues/4

- adds check for PR closures
- adds tests

I think there may be legitimate scenarios where a PR exists cross-repo but I'm unsure about cross-org. But that will likely be the exception rather than the rule, how do you feel about this approach?

```ts
for await (const pr of linkedPullRequests) {
    if (pr.author !== author || pr.organization !== currentRepo.owner.login || pr.repository !== currentRepo.name) {
      continue;
```

QA: https://github.com/ubq-testing/command-start-stop/issues/2#issuecomment-2228740992

- I have two opened PRs one by this account, one by my testing account and the timeline contains PRs from this repo. The comment shows only this account's PR being closed.


0x4007:
> I think there may be legitimate scenarios where a PR exists cross-repo but I'm unsure about cross-org. But that will likely be the exception rather than the rule

I think we should have support for all the above scenarios. Cross org indeed is rare. If its too complicated then perhaps we can make a new task for it (if needed.)

Keyrxng:
> Cross org indeed is rare. If its too complicated then perhaps we can make a new task for it (if needed in the future.)

I think a new task if required is the way to go. 

Correct me if I'm wrong, but for cross-org closures we'd need the bot to be installed in the other org right? I'm assuming based on the earlier attempts of my bot to close #1 and it never succeeding

Keyrxng:
I want to add that even with the example in the PR body it would still likely be possible for a PR to be closed in error within the same repo depending on what is being cross-referenced

until it's live-tested I can't be sure how effective it's going to be. We are making assumptions about what is being cross-referenced in production and I know there's a good chance it'll close a PR in error eventually with cross-repo closures.

Keyrxng:
I've added the PR body check

I referenced the issue from another repo from both an issue and PR. And I couldn't find either of those references in the timeline pre-filtering, which is odd right because previously it was showing PRs between this org and mine.

The sooner we see it live QA-d the better and I'm happy to handle any fixes/tweaks that are needed

fresh QA:

https://github.com/ubq-testing/command-start-stop/issues/5

.github/knip.ts
```diff
@@ -3,10 +3,10 @@ import type { KnipConfig } from "knip";
 const config: KnipConfig = {
   entry: ["build/index.ts"],
   project: ["src/**/*.ts"],
-  ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/worker.ts"],
+  ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/worker.ts", "src/types/plugin-input.ts", "src/main.ts", "src/plugin.ts"],
```
gentlementlegen, line 6: Why are these ignored?

src/handlers/shared/stop.ts
```diff
@@ -7,23 +7,17 @@ export async function stop(context: Context, issue: Context["payload"]["issue"],
...
   const assignees = issue.assignees ?? [];
-  if (assignees.length == 0) {
-    logger.error("No assignees found for issue", { issueNumber });
-    await addCommentToIssue(context, "````diff\n! You are not assigned to this task.\n````");
-    return { output: "No assignees found for this task" };
-  }
-
   // should unassign?
-  const shouldUnassign = assignees[0]?.login.toLowerCase() == sender.login.toLowerCase();
+  const whoToUnassign = assignees.find((assignee) => assignee?.login.toLowerCase() === sender.login.toLowerCase());
```
gentlementlegen, line 11:
```suggestion
  const usersToUnassign = assignees.find((assignee) => assignee?.login.toLowerCase() === sender.login.toLowerCase());
```

src/utils/issue.ts
```diff
@@ -169,3 +194,38 @@ async function getOpenedPullRequests(context: Context, username: string): Promis
...
+ * `Requires https://www.github.com/.../issue/123`
+ * `Fixes https://github.com/.../issues/123`
+ * `Closes #123`
+ * `Depends on #123`
+ * `Related to #123`
+ */
+export function issueLinkedViaPrBody(prBody: string | null, issueNumber: number): boolean {
+  const regex = // eslint-disable-next-line no-useless-escape
+    /(?:Resolves|Fixes|Closes|Depends on|Related to) #(\d+)|https:\/\/(?:www\.)?github.com\/([^\/]+)\/([^\/]+)\/(issue|issues)\/(\d+)|#(\d+)/gi;
+  const matches = prBody?.match(regex);
```
gentlementlegen, line 210: The problem I noticed with this, is that in our issue template we have a comment that contains `Resolves #1234`. This matches the regex as well because comments are not filtered out, which I believe is also the case here.

tests/main.test.ts
```diff
@@ -196,6 +240,7 @@ describe("User start/stop", () => {
   });
 
   test("User can't start another issue if they have reached the max limit", async () => {
+    // getAvailableOpenedPullRequests()
```
gentlementlegen, line 243:
```suggestion
```

.github/knip.ts
Keyrxng: https://github.com/ubiquibot/command-start-stop/pull/10#issuecomment-2228807316 this knip comment above

https://github.com/ubiquibot/command-start-stop/actions/runs/10049073859/job/27774413984?pr=10

> Unused item in ignoreDependencies: @actions/core
> Unused item in ignoreDependencies: @actions/github

These are the only ignored items that are not hitting apparently

src/handlers/shared/stop.ts
Keyrxng: Should we use a plural when it only returns a singular user? and given that `/stop` only unassigns the invoker and never others I think user singular is more appropriate

gentlementlegen: You are correct I put plural by accident, singular makes sense.

src/utils/issue.ts
Keyrxng: hmm, the `"pull_request" in source` check should eliminate any issues from the timeline afaik

But both of these failed to hit my timeline when I was QA-ing and I'm clueless as to why other than they are in a different repo. Unsure how to proof things better without plugging it into V2 for testing but that's not ideal

https://github.com/ubq-testing/bot-ai/pull/48
https://github.com/ubq-testing/bot-ai/issues/47

gentlementlegen: I meant that the `prBody` would be like:
```
Resolves #12

<!-- Resolves #1234 -->
```

Second one is a comment and should be ignored.

.github/knip.ts
gentlementlegen: Are these actually used?
Is knip actually running on this branch during the tests? I suspect it runs on the `development` branch to post this.

Keyrxng: the workflow runs on `pull_request_target` and the actions tab shows it pitted against my fork/branch so I believe it's running against this PR.

Yeah everything mentioned in the knip comment above is being used, perhaps not the `main.ts` and `plugin.ts`

src/utils/issue.ts
gentlementlegen: This issue also exists in other repos sadly.

Keyrxng: what do you suggest then, avoid using this sort of approach all together and revert back to my first approach or solidify it?

src/handlers/shared/stop.ts
```diff
@@ -7,23 +7,17 @@ export async function stop(context: Context, issue: Context["payload"]["issue"],
...
-  }
-
   // should unassign?
-  const shouldUnassign = assignees[0]?.login.toLowerCase() == sender.login.toLowerCase();
+  const userToUnassign = assignees.find((assignee) => assignee?.login.toLowerCase() === sender.login.toLowerCase());
 
-  if (!shouldUnassign) {
+  if (!whoToUnassign) {
     logger.error("You are not assigned to this task", { issueNumber, user: sender.login });
+    await addCommentToIssue(context, "```diff\n! You are not assigned to this task.\n```");
```
gentlementlegen, line 15: Isn't the logger supposed to also generate the diff?

Keyrxng: I forgot about that @gentlementlegen 

[QA](https://github.com/ubq-testing/command-start-stop/issues/2#issuecomment-2245996914)

src/utils/issue.ts
Keyrxng: Solved for html comments but even still I expect to see some teething problems with this especially with any of the selectors other than "Resolves"

.github/knip.ts
gentlementlegen: If the `main` is not recognized possibly some files will be considered as unused because they are reference through the main. But in other projects we never needed to ignore these files.

src/utils/issue.ts
gentlementlegen: I think a quick fix that would be reliable is to remove comment, by using a substring with a regex, so comments are ignored, then your logic should be good.

0x4007: I always open up my pull body with "Resolves #0". Are we sure that GitHub correctly links the pull if "Resolves" is located elsewhere? 

`^Resolves\s#\d+?\D`

gentlementlegen: It does resolve it properly wherever it is located, except HTML comments that are also properly ignored.

.github/knip.ts
Keyrxng: resolved

.github/knip.ts
```diff
@@ -6,7 +6,7 @@ const config: KnipConfig = {
   ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/worker.ts"],
   ignoreExportsUsedInFile: true,
   // eslint can also be safely ignored as per the docs: https://knip.dev/guides/handling-issues#eslint--jest
-  ignoreDependencies: ["eslint-config-prettier", "eslint-plugin-prettier", "@types/jest"],
+  ignoreDependencies: ["eslint-config-prettier", "eslint-plugin-prettier", "@types/jest", "@actions/core", "@actions/github"],
```
gentlementlegen, line 9: @Keyrxng So in the end are these needed?

Keyrxng: @gentlementlegen https://github.com/ubiquibot/command-start-stop/actions/runs/10077586980/job/27860439377?pr=10

Pretty sure they are needed yeah otherwise we'd see this warning in the above run log

> Unused item in ignoreDependencies: @actions/core
> Unused item in ignoreDependencies: @actions/github

gentlementlegen: @Keyrxng Yes but these dependencies are actually never imported anywhere in the codebase from what I see, so Knip message makes sense, they can be removed from `package.json`.

Keyrxng: Ah I was on auto-pilot thinking they were in the action entry but it's been removed

gentlementlegen (approved):

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
+// @ts-check
```
0x4007, line 1: Seems unusual that you added this directive. I suppose it doesn't automatically pick up mjs files?

Also why mjs? We don't normally use this format according to the template.

src/handlers/user-start-stop.ts
```diff
@@ -10,8 +10,8 @@ export async function userStartStop(context: Context): Promise<{ output: string
   const { isEnabled } = config;
```
0x4007, line 10: This enabled stuff should be removed.

0x4007:
After my comments are addressed I'm okay to merge.

eslint.config.mjs
gentlementlegen: This is the recommended format from [the docs](https://typescript-eslint.io/getting-started/#step-2-configuration). And indeed the `ts-check` enforces the same checks as the other files (would not by default since it is `js`).

This is now used everywhere when `eslint` was updated to the latest version, as it became the standard.

src/handlers/user-start-stop.ts
gentlementlegen: I guess this was to mimic the v1 that actually had a field to disable the start stop command, which I don't think is needed for v2.

Keyrxng: I disagree. How can we disable it on specific repos with this removed without adding a config item to disable it for repos?

gentlementlegen: Currently, the repo configuration will override the organization repo. There is an issue on that topic [here](https://github.com/ubiquity/ubiquibot-kernel/issues/54).

I would say that if in the target repository configuration, the plugin is not mentioned, it won't run (or be disabled if you will). If it is present, the target repo configuration will be used.

Keyrxng: > I would say that if in the target repository configuration, the plugin is not mentioned, it won't run (or be disabled if you will). If it is present, the target repo configuration will be used.

So it would be active globally via the org and if there is no config in the repo or there is but no mention of the plugin then it's disabled in that repo? I think I must misunderstand as that would mean it would be mandatory to define a repo config with this plugin wherever it should be "enabled"?

gentlementlegen: No. If it is enabled organization level, it will be enabled in every repo. I a repo defines a configuration with `command-start-stop` in it, it will override totally the organization configuration. If a repo defines a configuration without `command-start-stop` then it will never run on that repo.

Keyrxng: Right okay I'm following so far. 

So the pattern we've been using like `config?.item ? action() : null` is that still relevant for determining if a plugin should run? Most recently it happened in [xpCheck](https://github.com/ubiquibot/command-start-stop/pull/17/files#diff-1181dad3ddb1451203d2941f1b46bf3c3ea60bcc8e0da7093513931d0d11e518) but that's for a sub-feature of the plugin I guess, so would that pattern apply in similar scenarios for enabling/disabling specific features?

gentlementlegen: Yes, implemented [here for example](https://github.com/ubiquibot/conversation-rewards/pull/75#issuecomment-2253656887).

0x4007: > No. If it is enabled organization level, it will be enabled in every repo. I a repo defines a configuration with `command-start-stop` in it, it will override totally the organization configuration. If a repo defines a configuration without `command-start-stop` then it will never run on that repo.

Can you demonstrate how the config would look if you wanted to disable it on a repo?

gentlementlegen:
```yml
plugins:
    # any plugin you want except this one
```

0x4007: This doesn't seem right. Configuration merging should be at the resolution of the plugin id level. And I would have expected that in order to disable this we would have something like 
```yml
plugins:
   issues_comment.created:
      - id: ubiquibot/command-start-stop
        with: 
```

gentlementlegen:
@Keyrxng Yes, the merge issue is unrelated to this and will be addressed within the kernel. I am fine with this pull-request.

src/handlers/user-start-stop.ts
0x4007: > if you cannot see the configuration file.

The bot has access to private repositories. Using the bot authentication, you should be able to access the organization config.

PR #11, opened 2024-07-15
Feat/teams

### LINKED ISSUE #6: Teams `/start @user`
This reminds me that it would be really interesting in the future to support teams. Where I could write on a new task 

```
/start @Keyrxng 
```

And we would both be assigned (and be eligible to split the assignee reward.)

I know a lot of developers work as lone rangers but it could make the experience more collaborative and exciting if we have a massive tasks marketplace and teams hunt projects to co-op on together.

Also if I am already assigned to a project and I write the above command, then it will add them as an assignee as well.

Keyrxng:
Resolves #6
Requires #10

QA: https://github.com/ubq-testing/command-start-stop/issues/2#issuecomment-2228846308

### DIFF
--- a/src/handlers/shared/start.ts
+++ b/src/handlers/shared/start.ts
@@ -61,13 +61,23 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     throw new Error("Issue is closed");
   }
 
-  const assignees = (issue?.assignees ?? []).filter(Boolean);
+  const assignees = issue?.assignees ?? []
+
   if (assignees.length !== 0) {
     // const log = logger.error("The issue is already assigned. Please choose another unassigned task.", { issueNumber: issue.number });
     // await addCommentToIssue(context, log?.logMessage.diff as string);
     const currentUserAssigned = !!assignees.find((assignee) => assignee?.login === sender.login);
-    const log = logger.error(currentUserAssigned ? "You are already assigned to this task." : "The issue is already assigned. Please choose another unassigned task.", { issueNumber: issue.number });
-    return await addCommentToIssue(context, log?.logMessage.diff as string);
+    const comment = currentUserAssigned ? "You are already assigned to this task." : "The issue is already assigned. Please choose another unassigned task.";
+    await addCommentToIssue(context, `\`\`\`diff\n! ${comment}`);
+    throw new Error(comment);
+  }
+
+  teammates.push(sender.login)
+
+  // check max assigned issues
+  for (const user of teammates) {
+    if (!user) continue;
+    await handleTaskLimitChecks(user, context, maxConcurrentTasks, logger, sender.login);
   }
 
   // get labels
@@ -117,4 +127,20 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   );
 
   return { output: "Task assigned successfully" };
+}
+
+async function handleTaskLimitChecks(username: string, context: Context, maxConcurrentTasks: number, logger: Context["logger"], sender: string) {
+  const openedPullRequests = await getAvailableOpenedPullRequests(context, username);
+  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed: `, { openedPullRequests });
+
+  const assignedIssues = await getAssignedIssues(context, username);
+  logger.info("Max issue allowed is", { maxConcurrentTasks, assignedIssues: assignedIssues.map((issue) => `${issue.url}`) });
+
+  // check for max and enforce max
+  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
+    const isSender = username === sender;
+    const comment = (isSender ? "You have" : `${username} has`) + ` reached the max limit of ${maxConcurrentTasks} assigned issues.`;
+    await addCommentToIssue(context, `\`\`\`diff\n! ${comment}\n\`\`\``);
+    throw new Error(`Too many assigned issues, you have reached your max limit of ${maxConcurrentTasks} issues.`);
+  }
 }
--- a/src/handlers/shared/start.ts
+++ b/src/handlers/shared/start.ts
@@ -24,7 +24,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   let commitHash: string | null = null;
 
   try {
-    const hashResponse = await context.octokit.repos.getCommit({
+    const hashResponse = await context.octokit.rest.repos.getCommit({
       owner: context.payload.repository.owner.login,
       repo: context.payload.repository.name,
       ref: context.payload.repository.default_branch,
--- a/src/handlers/shared/start.ts
+++ b/src/handlers/shared/start.ts
@@ -1,4 +1,4 @@
-import { Assignee, Context, ISSUE_TYPE, Label } from "../../types";
+import { Context, ISSUE_TYPE, Label } from "../../types";
 import { isParentIssue, getAvailableOpenedPullRequests, getAssignedIssues, addAssignees, addCommentToIssue } from "../../utils/issue";
 import { calculateDurations } from "../../utils/shared";
 import { checkTaskStale } from "./check-task-stale";
@@ -34,25 +34,6 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     logger.error("Error while getting commit hash", { error: e as Error });
   }
 
-  // check max assigned issues
-  const openedPullRequests = await getAvailableOpenedPullRequests(context, sender.login);
-  logger.info(`Opened Pull Requests with approved reviews or with no reviews but over 24 hours have passed: `, { openedPullRequests });
-
-  const assignedIssues = await getAssignedIssues(context, sender.login);
-  logger.info("Max issue allowed is", { maxConcurrentTasks, assignedIssues: assignedIssues?.map((issue) => `${issue.url}`) });
-
-  // check for max and enforce max
-
-  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
-    const log = logger.error("Too many assigned issues, you have reached your max limit", {
-      assignedIssues: assignedIssues.length,
-      openedPullRequests: openedPullRequests.length,
-      maxConcurrentTasks,
-    });
-    await addCommentToIssue(context, log?.logMessage.diff as string);
-    throw new Error(`Too many assigned issues, you have reached your max limit of ${maxConcurrentTasks} issues.`);
-  }
-
   // is it assignable?
 
   if (issue.state === ISSUE_TYPE.CLOSED) {
@@ -63,6 +44,7 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
 
   const assignees = issue?.assignees ?? [];
 
+  // find out if the issue is already assigned
   if (assignees.length !== 0) {
     const isCurrentUserAssigned = !!assignees.find((assignee) => assignee?.login === sender.login);
     const log = logger.error(
@@ -77,7 +59,6 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
 
   // check max assigned issues
   for (const user of teammates) {
-    if (!user) continue;
     await handleTaskLimitChecks(user, context, maxConcurrentTasks, logger, sender.login);
   }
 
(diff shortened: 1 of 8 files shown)


Keyrxng:
@0x4007 Should I impose assignedIssue checks on both teammates or just the command invoker?

0x4007:
> @0x4007 Should I impose `maxConcurrentTasks` checks on all teammates or just the command invoker?

I don't really understand the full context of the question. I think in the future this makes more sense to discuss under the issue specification instead.

`maxConcurrentTasks` is a setting that allows us to let some users only work on up to a certain amount of tasks at any given time. 

Core team members should have higher limits compared to external people. 

The check only happens upon task assignment `/start`

Keyrxng:
given that this PR adds extends `/start` to allow assigning more than one person, should those checks be done against only the command invoker or should we check each user in the command string?

If done only against the invoker users could get around these checks but doing it against all might limit collaboration.

Currently the check is performed against the command invoker only

0x4007:
> Currently the check is performed against the command invoker only

> check each user

Checking each user seems correct based on the vision.

src/handlers/shared/start.ts
```diff
@@ -83,16 +94,29 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
   const assignmentComment = await generateAssignmentComment(context, issue.created_at, issue.number, id, duration);
   const metadata = structuredMetadata.create("Assignment", logMessage);
 
-  // add assignee
-  if (!assignees.map((i: Partial<Assignee>) => i?.login).includes(login)) {
-    await addAssignees(context, issue.number, [login]);
+  const toAssign = [];
+
+  for (const teammate of teammates) {
+    if (!assignees.find((assignee: Partial<Assignee>) => assignee?.login?.toLowerCase() === teammate.toLowerCase())) {
```
whilefoo, line 112: before you already checked that `assignees` is empty, no?

src/handlers/shared/start.ts
```diff
@@ -111,3 +135,19 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
 
   return { output: "Task assigned successfully" };
 }
+
+async function handleTaskLimitChecks(username: string, context: Context, maxConcurrentTasks: number, logger: Context["logger"], sender: string) {
```
whilefoo, line 139: isn't this code the same as in start function? then you can call this function instead of repeating code

whilefoo (changes requested):

Keyrxng:
QA: https://github.com/ubq-testing/bot-ai/issues/27#issuecomment-2271214304

There seems to be a bug that I cannot seem to resolve. In the QA issue no problems, it's a public repo and public tasks. I also have a private issue in my `ubiquibot-config` repo and for some reason it just will not let me assign two users at once.

I can self assign as `keyrxng2` but I cannot dual-assign from either account. I am logging the assignees and both are being passed, nothing is different logic wise just the repo/issue env.

image

I am passing the array to the `octokit.addAssignees()` and when I use a for-of loop I get this rapid assigning/unassigning
image

---

Have you folks encountered anything like this before?

whilefoo:
> I can self assign as `keyrxng2` but I cannot dual-assign from either account

I'm not completely sure but I think that's because for private repos you need Github subscription to have that option

Keyrxng:
> > I can self assign as `keyrxng2` but I cannot dual-assign from either account
> 
> I'm not completely sure but I think that's because for private repos you need Github subscription to have that option

https://github.com/orgs/community/discussions/7150

Seems you are correct @whilefoo. We'd need permissions to access billing to determine this I'm sure so if we are to implement some sort of custom message then would we do it by simply checking `assignees.length` after assigning? 

Or no custom message and deal with it if we need to later?

src/utils/issue.ts
```diff
@@ -8,19 +8,14 @@ export function isParentIssue(body: string) {
...
-      ({ data: issues }) => issues.filter((issue: Issue) => !issue.pull_request && issue.assignee && issue.assignee.login === username)
-    );
+    return (await context.octokit.search
+      .issuesAndPullRequests({
+        q: `is:open assignee:${username} org:${payload.repository.owner.login}`,
+      })
+      .then((response) => response.data.items)) as Issue[];
   } catch (err: unknown) {
     context.logger.error("Fetching assigned issues failed!", { error: err as Error });
     return [];
```
whilefoo, line 21: throw error

src/utils/issue.ts
```diff
@@ -8,19 +8,14 @@ export function isParentIssue(body: string) {
...
-        owner: payload.repository.owner.login,
-        repo: payload.repository.name,
-        state: ISSUE_TYPE.OPEN,
-        per_page: 100,
-      },
-      ({ data: issues }) => issues.filter((issue: Issue) => !issue.pull_request && issue.assignee && issue.assignee.login === username)
-    );
+    return (await context.octokit.search
+      .issuesAndPullRequests({
+        q: `is:open assignee:${username} org:${payload.repository.owner.login}`,
```
whilefoo, line 16:
```suggestion
        q: `is:open is:issue assignee:${username} org:${payload.repository.owner.login}`,
```
to match only issues

src/utils/issue.ts
```diff
@@ -8,19 +8,14 @@ export function isParentIssue(body: string) {
...
-      {
-        owner: payload.repository.owner.login,
-        repo: payload.repository.name,
-        state: ISSUE_TYPE.OPEN,
-        per_page: 100,
-      },
-      ({ data: issues }) => issues.filter((issue: Issue) => !issue.pull_request && issue.assignee && issue.assignee.login === username)
-    );
+    return (await context.octokit.search
+      .issuesAndPullRequests({
```
whilefoo, line 15: consider adding paginate because even though it will still work because we just want to know if user has issues or not, somebody else might use this function thinking it will return all issues.

whilefoo (changes requested):

whilefoo:
> Or no custom message and deal with it if we need to later?

I think it's fine without the messages (most orgs will have paid plans)

src/utils/issue.ts
Keyrxng: My mistake I just double-checked the [docs](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests) too to correct myself, I used the UI to test the query and must have been in the wrong list as it threw an error so I avoided it here

That endpoint is so OP it's crazy

0x4007:
> > Or no custom message and deal with it if we need to later?
> 
> 
> 
> I think it's fine without the messages (most orgs will have paid plans)

If this isn't a lot of work I think just include the custom error message within this pull? Seems to me like this can be handled in a few lines of code.

src/utils/issue.ts
```diff
@@ -112,11 +104,42 @@ export async function closePullRequestForAnIssue(context: Context, issueNumber:
...
+    await addCommentToIssue(context, log?.logMessage.diff as string);
+  }
+}
+
 export async function addAssignees(context: Context, issueNo: number, assignees: string[]) {
   const payload = context.payload;
 
   try {
-    await context.octokit.rest.issues.addAssignees({
+    await context.octokit.issues.addAssignees({
```
gentlementlegen, line 142: Probably should use `rest`

Keyrxng: I thought that was purely a preference thing since `octokit` wraps all of the modules but I see the pros of using `rest`, will do this from now on

gentlementlegen: Had a talk with @whilefoo and `rest` seems to be the modern way to use it, the other one being the legacy.

src/utils/issue.ts
```diff
@@ -107,6 +104,37 @@ export async function closePullRequestForAnIssue(context: Context, issueNumber:
...
+  } = await octokit.rest.issues.get({
+    owner: payload.repository.owner.login,
+    repo: payload.repository.name,
+    issue_number: issueNumber,
+  });
+
+  if (!assignees?.length) {
+    const log = logger.error("We detected that this task was not assigned to anyone. Please report this to the maintainers.", { issueNumber, usernames });
+    await addCommentToIssue(context, log?.logMessage.diff as string);
+    throw new Error(log?.logMessage.raw);
```
whilefoo, line 127: consider throwing the error which `logger.error` returns and let it bubble up and add a catch at the start of the chain where it posts a comment
for reference [this](https://github.com/ubiquibot/conversation-rewards/blob/development/src/index.ts) is how it's done in `conversation-rewards`

whilefoo (changes requested):

src/utils/issue.ts
Keyrxng: So in this async-await context would that just look like a try-catch at the beginning of this function and `addCommentToIssue` from the catch block?

The reference is in a workflow context, so replicating that as-is we'd wrap `runPlugin()` in `plugin.ts` and we bubble up all errors from across the logic and post from there?

Keyrxng: @whilefoo I have placed it in `plugin.ts` wrapping the `userStartStop()` function. The reason why is because if we do it inside `worker.ts` then we need to rebuild `context` to pass into `addCommentToIssue` otherwise split `worker.ts` so that it's globally accessible from the catch block which means two try blocks one for everything before `await request.json()` and one for everything after. 

It seems like a smoother approach handling it after `context` is built and easily accessible however I can refactor `worker.ts` if you'd prefer that way

tests/main.test.ts
```diff
@@ -110,22 +110,53 @@ describe("User start/stop", () => {
...
+        user: "ubiquity",
+      },
+    };
 
-    expect(output).toEqual({ output: "You are not assigned to this task" });
+    try {
+      expect(await userStartStop(context as unknown as Context)).toThrow(logReturn);
+    } catch (error) {
+      expect(error).toEqual(logReturn);
+    }
```
whilefoo, line 131: because you awaited the function inside expect, it will return a value so `toThrow` is meaningless, but it will still work because you catch the error.

I recommend:
```suggestion
    await expect(userStartStop(context as unknown as Context)).rejects.toThrow(logReturn)
```

whilefoo (changes requested):

tests/main.test.ts
Keyrxng: Thank you, I was scratching my head as to why it wasn't working like this. I was getting a ts error putting `await` before expect but perhaps it was because I had awaited the function too.

Keyrxng: Unsure what the issue is but your suggestion is not working as expected.

without the try-catch it tells me that the function did not throw but you see in the screenshot below inside the try-catch it's displaying our `LogReturn` that we throw... Seen anything like this before?

image

whilefoo: I'm not sure but it's in the [docs](https://jestjs.io/docs/expect#rejects) so it should work

Keyrxng: the issue seemed to be that I wasn't throwing an `Error` so it wasn't registering

src/plugin.ts
```diff
@@ -22,7 +23,17 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
...
+    try {
+      return await userStartStop(context);
+    } catch (err) {
+      let errorMessage;
+      if (err instanceof LogReturn) {
+        errorMessage = context.logger.error(`Failed to run comment evaluation. ${err.logMessage?.raw || err}`, { err });
+      } else {
+        errorMessage = context.logger.error(`Failed to run comment evaluation. ${err}`, { err });
+      }
+      await addCommentToIssue(context, `${errorMessage?.logMessage.diff}\n<!--\n${JSON.stringify(errorMessage?.metadata, null, 2)}\n-->`);
```
gentlementlegen, line 35: It is a bit risky to stringify the object without making sure that the content is not escaped. If there is any special character (`<`, `>`, `-->` etc) there is a change the whole comment break. It would be great to sanitize it before.

src/plugin.ts
```diff
@@ -32,9 +32,56 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
...
   } else {
     context.logger.error(`Unsupported event: ${context.eventName}`);
   }
 }
+
+function sanitizeDiff(diff?: LogReturn["logMessage"]["diff"]): string {
+  if (!diff) return "";
+  // eslint-disable-next-line no-useless-escape
+  const backticks = diff.match(/\`\`\`/g);
+  if (!backticks) return "```\n" + diff + "\n```";
```
whilefoo, line 47: why do we need to add ```, the diff already has that?

Keyrxng: I know I thought this myself but when gentlementlegen mentioned that the comment would break I realised it would also break if somehow there was only set of backticks although I do agree it's pretty pointless.

Agree to remove?

0x4007: @gentlementlegen rfc

Keyrxng: I want to add in case it's not clear, I mean to remove me sanitizing the diff as it's very likely not needed but the metadata object sanitizing should remain

gentlementlegen: If the backticks are not escaped they indeed could probably break the comment. It would be safer to transform them to their related code `&#96` if you intend to include them in GitHub comments, which would be safer. Otherwise our logger should take care of adding these around the comment body.

Keyrxng: escaping them breaks formatting when posted, edit this message to see. I think we could just remove it because the logger is guaranteed to diff it correctly. The only way I could break formatting is if a set of backticks between between two new lines exists which I don't think is possible from anything that we log in the initial string. So I'll remove it.

&#96;&#96;&#96;diff
! Failed to run comment evaluation.
&#96;&#96;&#96;

gentlementlegen (approved):
Good to go for me when you addressed the remaining comments.

whilefoo (approved):

PR #13, opened 2024-07-22
Disable Self Assigns

### LINKED ISSUE #2: Check For Unassigns; Disable Self Assigns
> Before disqualification, the manager should ask for an update. If the contributor still has no updates, then the manager should disqualify the contributor by unassigning them, and ideally, not allowing them to reassign themselves (we could check in the `/start` command if the contributor was ever unassigned within the issue timeline.)

_Originally posted by @0x4007 in https://github.com/ubiquibot/plugins-wishlist/issues/7#issuecomment-2146804481_

jordan-ae:
Resolves #2

### DIFF
--- a/src/handlers/shared/check-unassigns.ts
+++ b/src/handlers/shared/check-unassigns.ts
@@ -1,0 +1,20 @@
+import { Context } from "../../types";
+
+export async function wasPreviouslyUnassigned(
+  context: Context,
+  senderLogin: Context["payload"]["sender"],
+  issueNumber: Context["payload"]["issue"]
+): Promise<boolean> {
+  try {
+    const issueEvents = await context.octokit.issues.listEventsForTimeline({
+      owner: context.payload.repository.owner.login,
+      repo: context.payload.repository.name,
+      issue_number: issueNumber.id,
+    });
+
+    return issueEvents.data.some((event) => event.event === "unassigned" && event.event === senderLogin);
+  } catch (error) {
+    console.error("Error while checking previous assignment:", error);
+    return false;
+  }
+}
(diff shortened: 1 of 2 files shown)


0x4007:
Code looks good. Can you prove it works as expected with a QA test?

gentlementlegen:
I would like to see associated tests and a QA run somewhere. Also, maybe this should be configurable through the plugin configuration if we want users to be able to assign themselves again or not.

src/handlers/shared/check-unassigns.ts
```diff
@@ -0,0 +1,20 @@
...
+  issueNumber: Context["payload"]["issue"]
+): Promise<boolean> {
+  try {
+    const issueEvents = await context.octokit.issues.listEventsForTimeline({
+      owner: context.payload.repository.owner.login,
+      repo: context.payload.repository.name,
+      issue_number: issueNumber.id,
+    });
+
+    return issueEvents.data.some((event) => event.event === "unassigned" && event.event === senderLogin);
```
Keyrxng, line 15: I'm sceptical that `event.event` can be both `unassigned` and `senderLogin`, QA would be ideal

src/handlers/shared/check-unassigns.ts
```diff
@@ -0,0 +1,20 @@
+import { Context } from "../../types";
+
+export async function wasPreviouslyUnassigned(
+  context: Context,
+  senderLogin: Context["payload"]["sender"],
+  issueNumber: Context["payload"]["issue"]
```
Keyrxng, line 6: naming convention for these args is off slightly so it's being used in error in the return statement below I think

jordan-ae:
@gentlementlegen &  @Keyrxng this isn't ready for reviews yet. I sent it in so I can save my progress that's why I left it on draft. @0x4007 you should probably revert your approval 😅😅

Keyrxng:
Normally I wouldn't have reviewed but it was because there was an approval on it I checked it too, I assumed you were waiting till you get the other one through first

src/handlers/shared/check-unassigns.ts
0x4007: Just noticed that this fails on the test CI. I suppose I should be more cautious of reviews from mobile as it is not easy to see these details sometimes.

0x4007: Just noticed that this fails on the test CI. I suppose I should be more cautious of reviews from mobile as it is not easy to see these details sometimes.

0x4007 (changes requested):

PR #17, opened 2024-07-27
Feat/experience check

Keyrxng:
Resolves https://github.com/ubiquibot/plugins-wishlist/issues/26

It's rudimentary but something is better than nothing at least and is a good starting point for extending things further

- Uses https://github-readme-stats.vercel.app [`repo`](https://github.com/anuraghazra/github-readme-stats) to pull stats as seen in my [readme](https://github.com/Keyrxng/Keyrxng)
- Adds `experience` config item

```ts
default: {
          minAccountAgeInDays: 365,
          mostImportantLanguage: { Typescript: 10 },
          languages: { Solidity: 10, JavaScript: 10 },
          statThresholds: {
            stars: 1,
            minCommitsThisYear: 1,
            prs: 1,
            issues: 1,
          },
        },
```

I'm unsure about default values other than `minAccountAgeInDays` and what ones we should give more weight to considering what we have to work with.

Rather than having to define a custom config for each repo I'm thinking we could pull stats using the `repo/statistics` endpoint for the payload repo and determine what sort of codebase it is then map languages to specific setups for `mostImportantLanguage` etc. But I think this should be handled in a separate PR.

---

If this would be better served as a standalone plugin let me know but it seems like this functionality should be built into the `start/stop` plugin.

QA:
- https://github.com/ubq-testing/command-start-stop/issues/8


gentlementlegen:
It is not written in the spec but I think there might be details on the requirements needed for the task displayed somewhere. Also given your code it seems to be based on the configuration, not task based. I think we should be able to set these individually per issue as it was stated in the specification.

Keyrxng:
> I think we should be able to set these individually per issue as it was stated in the specification.

This would require a change to how tasks are labelled to be effective wouldn't it and would also need a tier system that would correlate to the actual XP system that's not implemented at the moment.

> It would be great to know that collaborator solving the above issues has prior experience with solidity and ethereum.

The spec aims to guard Solidity tasks based on prior experience which is handled with this implementation effectively as anyone with that experience should have it in their language stats

My thinking is that any solidity based tasks would exist in solidity-heavy repos so defining a config in that repo with `mostImportantLanguage.solidity = 25` would suffice. Later maybe the config maps setups for each language eliminating the need for a custom config in each repo

gentlementlegen:
> This would require a change to how tasks are labelled to be effective wouldn't it
Yes that was what I thought, this way we could guard per issue, and also have any custom tag that would cover any tech (say we want Supabase experience, TypeScript, Solidity and whatnot). Otherwise I guess the finer we can get with a configuration is on a repo basis, which is maybe fine for now.

I do not know how important this is, and how fine grained we want to get! Maybe some input from @0x4007 would be helpful.

Keyrxng:
> Otherwise I guess the finer we can get with a configuration is on a repo basis, which is maybe fine for now.

I guess the markers other than account age could be removed as they seem sort of irrelevant and replaced with something like

```yml
experience:
 thresholds:
  solidity: 25
  typescript: 15
 languages:
  solidity:
   - ubiquity/ubiquity-dollar
   - etc..
  typescript:
   - pay.ubq.fi
   - etc...
```

rndquu:
I agree that we should be able to apply the "experience check" feature to individual github issues instead of the whole organization at once. 

We should somehow let the plugin know that the issue is guarded:
- either by a new label (like `Guard: solidity`)
- either by a comment in the issue specification (basically the same `Guard: solidity`). 

The issue with this approach is that it can be applied only to programming languages (keeping in mind that in the end we match contributor's languages from https://github.com/anuraghazra/github-readme-stats and original issue repository languages). It doesn't "guard" `UI/UX` or `devops` issues. Perhaps it's enough for v1.

0x4007:
A label that says `requires: solidity` could be intuitive and dynamically parsable.

Keyrxng:
> A label that says `requires: solidity` could be intuitive and dynamically parsable.

Okay so a configurable array of suffixes that prefix `requires: `, got it.

So say we apply the label, then we check the user's solidity percentage and what do we compare it against? Where does our issue-level threshold come from?

0x4007:
> Or we have `requires: Solidity (Easy)` etc, and map these to org-level customizable values?

Interesting idea!

`Required Experience: TypeScript (Senior)`
`Required Experience: TypeScript (Intermediate)`
`Required Experience: TypeScript (Junior)`
`Required Experience: TypeScript (Beginner)`

Keyrxng:
> Interesting idea! I think senior and junior are precise enough for our needs, and easily understandable.

So would this be a global org config item that contains all of our mappings to our label languages and tags etc? I thought we'd define everything within the plugin config like:

```yml
- with:
    labels: 
      - solidity:
          - junior: 5
          - mid: 15
          - pro: 50
          - rockstar: 70
```

or something like

```yml
- with:
    ranks:
      - junior: 5
      - mid: 15
    labels: ["Typescript"]
```

0x4007:
Let's keep the config simple and just follow [this strategy](https://github.com/ubiquibot/plugins-wishlist/issues/26#issuecomment-2259961394).

Keyrxng:
Closed because this will be it's own plugin and not an extension of `start-stop`, I'll pick this up tomorrow

Keyrxng:
https://github.com/ubiquibot/task-xp-guard/pull/1

PR #19, opened 2024-07-28
feat: disabled multi self assign

### LINKED ISSUE #2: Check For Unassigns; Disable Self Assigns

Keyrxng:
Resolves #2

- adds `botUsernames` into the config as I think each partner's bot will have a different name
- handles checks against admin, comment and github assignments
- added simple test case

Should this also listen for a new event to detect if they are assigning via the UI? It's also needed for https://github.com/ubiquibot/command-start-stop/issues/12 so I think it makes sense?


Keyrxng:
Shoot I meant to PR this against jordan's PR. Does it make any difference in the end @gentlementlegen?

gentlementlegen:
@Keyrxng If it contains the other PR content no it doesn't. However I think that the other pull-request comments won't be evaluated. No impact on the reward itself.

Keyrxng:
> @Keyrxng If it contains the other PR content no it doesn't. However I think that the other pull-request comments won't be evaluated. No impact on the reward itself.

I understand so it's not really a concern here as the bot likely wouldn't have credited much in relevance for the one contributor comment anyway

gentlementlegen:
@Keyrxng Correct. I will review this pull-request instead then.

Keyrxng:
I just thought we could add a configurable limit to how many times a user can be unassigned and since we have three ways to track i.e admin, ui or self we could have limits for them individually if it makes sense to do that

README.md
```diff
@@ -43,6 +43,8 @@ To configure your Ubiquibot to run this plugin, add the following to the `.ubiqu
     miscellaneous:
       maxConcurrentTasks: 3
       startRequiresWallet: true # default is true
+      botUsernames:
+        - ubiquibot[bot]
```
gentlementlegen, line 47: What is it for?

Keyrxng: I'm best guessing that our bot name won't be used across all partner's installs so it's needed to allow them to filter their own bot name

image

Keyrxng: I thought just use `.type === "Bot"` but felt this might be the preferred approach

gentlementlegen: I think for now using `Bot` is okay because it is easy to make a typo or forget about this field. Having the two bots running is a case that should not be happening often, and actually both bots should able not to clash with themselves. I suggest using `Bot` for now and if needed we can start to add extra configuration parameters as you did here.

0x4007:
> * adds `botUsernames` into the config as I think each partner's bot will have a different name

No we will run the official instance. Technically people can "self host" but in practice people generally pay for convenience. I don't think we should make special accommodations for self hosted instances.

README.md
0x4007: I think its a lot more robust to use the ID not the name. Besides we have at least one planned name change coming up soon i.e. UbiquityOS Superintendent / Kernel

Keyrxng: So you'd rather I use a an ID approach or just remove this config option completely and we'll hardcode our bot info?

src/handlers/shared/start.ts
```diff
@@ -21,6 +22,12 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     throw new Error("Issue is a parent issue");
   }
 
+  const hasBeenPreviouslyUnassigned = await checkPreviousAssignments(context, sender);
+  if (hasBeenPreviouslyUnassigned) {
+    await addCommentToIssue(context, "```diff\n! You were previously unassigned from this task. You cannot reassign yourself.\n```");
```
gentlementlegen, line 27: Might be better to create the comment with the logger.

README.md
0x4007: To be honest I don't feel that I have sufficient context because I can't even see the README.md in the code changes view. What problem are you trying to solve?

Keyrxng: Basically it's very likely that partners have other github bots running not just ours, I was trying to restrict the scope so we parse only assignment events from our bot

0x4007: Then you should identify our bot only by its user ID number.

tests/main.test.ts
```diff
@@ -279,6 +272,21 @@ describe("User start/stop", () => {
...
+    const issue = db.issue.findFirst({ where: { id: { equals: 2 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 2 } } }) as unknown as Sender;
+
+    const context = createContext(issue, sender, "/start");
+
+    try {
+      await userStartStop(context as unknown as Context);
+    } catch (error) {
+      if (error instanceof Error) {
+        expect(error.message).toEqual("This user was unassigned from this task previously. Cannot auto assign");
```
whilefoo, line 286: I think it's better to use
```js
await expect(userStartStop(context as unknown as Context)).rejects.toThrow('This user was unassigned from this task previously. Cannot auto assign');
```
because currently if the error is not thrown or error is not Error, the test will just end and you won't now something's wrong. 
also fix this in other places

src/handlers/shared/check-assignments.ts
```diff
@@ -0,0 +1,65 @@
...
+      });
+
+    return events
+      .filter((event) => event !== undefined)
+      .sort((a, b) => {
+        return new Date(a.createdAt || "").getTime() - new Date(b.createdAt || "").getTime();
+      });
+  } catch (error) {
+    context.logger.error("Error while getting assignment events", { error: error as Error });
+    return [];
```
whilefoo, line 63: it would be better to throw an error

src/handlers/shared/check-assignments.ts
```diff
@@ -0,0 +1,65 @@
...
+  const userAssignments = events.filter((event) => event.assignee?.toLowerCase() === senderLogin);
+
+  if (userAssignments.length === 0) {
+    return false;
+  }
+
+  const unassignedEvents = userAssignments.filter((event) => event.event === "unassigned");
+  const botUnassigned = unassignedEvents.filter((event) => event.actorType === "Bot");
+  const adminUnassigned = unassignedEvents.filter((event) => event.actor?.toLowerCase() !== senderLogin && event.actorType === "User");
+  const userSelfUnassignViaUi = unassignedEvents.filter((event) => event.actor?.toLowerCase() === senderLogin);
```
whilefoo, line 15: if an user self unassigns I don't think we should stop him from assigning again.

src/handlers/shared/check-assignments.ts
```diff
@@ -0,0 +1,65 @@
+import { Context, IssueEvent } from "../../types";
+
+export async function checkPreviousAssignments(context: Context, sender: Context["payload"]["sender"]): Promise<boolean> {
+  const events = await getAssignmentEvents(context);
+  const senderLogin = sender.login.toLowerCase();
```
whilefoo, line 5: I'm not sure if `sender` is always the user who created the comment, I think it's better to use `payload.comment.user.login` - this should be fixed across the whole codebase

src/handlers/shared/check-assignments.ts
```diff
@@ -0,0 +1,65 @@
+import { Context, IssueEvent } from "../../types";
+
+export async function checkPreviousAssignments(context: Context, sender: Context["payload"]["sender"]): Promise<boolean> {
```
whilefoo, line 3: since this function returns a boolean, the name is a bit weird. maybe something like `hasUserBeenUnassigned` or something similar

whilefoo (changes requested):

tests/main.test.ts
Keyrxng: I recently realized this with others working on the tests it's not ideal and I will avoid it future for sure

src/handlers/shared/check-assignments.ts
Keyrxng: between the two `payload.comment.user.login` is the only one typed as the user potentially being `undefined` so I've used both opting for `comment.user` first

src/handlers/shared/check-assignments.ts
```diff
@@ -0,0 +1,59 @@
+import { Context, IssueEvent } from "../../types";
+
+const UBIQUIBOT_ID = 113181824;
```
gentlementlegen, line 3: What does this value correspond to, the `APP_ID`? I think this should not be hardocded.

Keyrxng: https://github.com/ubiquibot/command-start-stop/pull/19#discussion_r1696328279

If it's configurable that allows partners to decide which by @0x4007's comments here that doesn't sound right

gentlementlegen: Got it, but then when we use the plugin within another organization won't this be broken?

Keyrxng: as far as I understand `user.id` will always be the same for the production Ubiquibot and won't ever change but it's name can change etc and `installID` would be different for each or

gentlementlegen: Then that would be fine I believe. My concern is when we are installing in a different organization so it doesn't break there, such as when we run our own tests.

Keyrxng: Ah I see yeah, for testing on V2 this wouldn't work. Perhaps I can hardcode our production and development botIDs and have a config item to use the dev bot?

tests/main.test.ts
```diff
@@ -258,34 +188,45 @@ describe("User start/stop", () => {
...
+    context.adapters = createAdapters(getSupabase(), context);
 
-    try {
-      await userStartStop(context as unknown as Context);
-    } catch (error) {
-      if (error instanceof Error) {
-        expect(error.message).toEqual("Too many assigned issues, you have reached your max limit of 3 issues.");
-      }
-    }
+    await userStartStop(context);
```
whilefoo, line 197: did you forget to check for error? and above too?

whilefoo (changes requested):

src/utils/issue.ts
```diff
@@ -198,7 +195,9 @@ export async function getAvailableOpenedPullRequests(context: Context, username:
 
   for (let i = 0; i < openedPullRequests.length; i++) {
     const openedPullRequest = openedPullRequests[i];
-    const reviews = await getAllPullRequestReviews(context, openedPullRequest.number);
+    const owner = openedPullRequest.html_url.split("/")[3];
+    const repo = openedPullRequest.html_url.split("/")[4];
```
whilefoo, line 199: consider creating a helper function that parses the url and also checks for array length

src/handlers/shared/start.ts
```diff
@@ -42,73 +33,69 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
   const labels = issue.labels;
   const priceLabel = labels.find((label: Label) => label.name.startsWith("Price: "));
 
   if (!priceLabel) {
-    throw logger.error("No price label is set to calculate the duration", { issueNumber: issue.number });
+    throw new Error(logger.error("No price label is set to calculate the duration", { issueNumber: issue.number }).logMessage.raw);
   }
 
   const duration: number = calculateDurations(labels).shift() ?? 0;
+  const toAssignIds = toAssign.map(async (u) => await fetchUserId(context, u));
```
whilefoo, line 86: check if any ids are null because that means the user doesn't exist (maybe we should throw error if that happens?)

src/handlers/shared/start.ts
```diff
@@ -129,16 +116,34 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
+
 async function handleTaskLimitChecks(username: string, context: Context, maxConcurrentTasks: number, logger: Context["logger"], sender: string) {
   const openedPullRequests = await getAvailableOpenedPullRequests(context, username);
   const assignedIssues = await getAssignedIssues(context, username);
 
   // check for max and enforce max
-  if (assignedIssues.length - openedPullRequests.length >= maxConcurrentTasks) {
-    throw logger.error(username === sender ? "You have reached your max task limit" : `${username} has reached their max task limit`, {
+
+  if (Math.abs(assignedIssues.length - openedPullRequests.length) >= maxConcurrentTasks) {
```
whilefoo, line 134: won't this mean if someone has 5 PRs but 2 issues assigned, they can't start a new task? I guess this is open for debate if they should be allowed to start a new task

whilefoo (changes requested):

src/handlers/shared/start.ts
0x4007: The original implementation allowed for productive contributors to work on two tasks concurrently:
- If any reviewer approved the pull, +1 
- If reviewers take longer than 24 hours, +1 
- If any reviewer requested changes, -1

Keyrxng: 2 (assigned) - 5 (PRs) = 3 === < 5 (maxTasks), is my understanding of that line of code. So someone with 0 assignedIssues and 5 PRs cannot be assigned anymore tasks.

> I'll make another task for this

To refactor this whole feature, so this requested change can be resolved for now?

0x4007: > We should also do +1 task if the PR author made necessary changes requested by the reviewer but are waiting for reviewer's next review and it's been more than 24 hours.
> 
> 
> 
> Currently the code fetches reviews but that endpoint returns all reviews even multiple reviews from the same person, so we need use only the last review by the same person but also check requested reviews. If a reviewer has requested changes but is being requested for review more than 24 hours, it should be considered +1 task.
> 
> 
> 
> I'll make another task for this

This is tricky. People generally aren't very discipline about changing it to a draft when they are continuing to edit it. What signal will be used to indicate that they are finished? We also can't always rely on them to formally request a new review from the UI. Most contributors will just tag the reviewers to check again. 

I don't think this is going to be viable.

src/handlers/shared/start.ts
Keyrxng: I went with throwing if null

gentlementlegen:
The README.md does not match the new configuration. Also, `APP_ID` seems to always be required in the configuration, is that something we want?

src/handlers/shared/start.ts
```diff
@@ -83,7 +83,15 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
...
-  const toAssignIds = toAssign.map(async (u) => await fetchUserId(context, u));
+  const toAssignIds = toAssign.map(async (u) => {
+    const userId = await fetchUserId(context, u);
+
+    if (!userId) {
+      throw new Error(logger.error("User not found", { user: u }).logMessage.raw);
+    }
+
+    return userId;
+  });
```
whilefoo, line 94: I've just realised that this will return an array of promises, you have to call `Promise.all()` to get results

src/worker.ts
```diff
@@ -36,6 +36,17 @@ export default {
...
+      if (!envConfigValidator.test(env)) {
+        const errorDetails: string[] = [];
+        for (const error of envConfigValidator.errors(env)) {
+          errorDetails.push(`${error.path}: ${error.message}`);
+        }
+        return new Response(JSON.stringify({ error: `Bad Request: the environment is invalid. ${errorDetails.join("; ")}` }), {
+          status: 400,
+          headers: { "content-type": "application/json" },
+        });
+      }
```
whilefoo, line 48: this will just validate it but the transformation won't be done. you need to use `Value.Decode` like above

whilefoo (changes requested):

whilefoo:
> Also, `APP_ID` seems to always be required in the configuration, is that something we want? Can't we deduce it from the bot app id itself if not set?

it's tedious for plugin developers to set that so I think we can include it in the SDK. I've looked at the webhook payload but I didn't find any information about app id

gentlementlegen:
`APP_ID` represents the installation id of the app within GitHub if I am not mistaken. It is called `UBIQUIBOT_APP_ID` in our org. Also checked the payload and this info is not there. And since this is not an action but a worker I guess it doesn't have the running context of GitHub so cannot figure it out by itself, probably.

whilefoo:
> `APP_ID` represents the installation id of the app within GitHub if I am not mistaken

I think it represents Github App's unique ID, installation ID is per org/repo

Keyrxng:
So nothing we can do with `APP_ID` the plugin dev just needs to pull this from their kernel worker `dev.vars` which they'd have to have running anyway so I don't think it's too much of an issue

gentlementlegen (approved):

whilefoo (approved):

PR #21, opened 2024-08-13
Refactor/time strings

Keyrxng:
Resolves #14

> Tasks to be carried out:
> - related tests

I'm assuming this will be merged quickly and once it and #9 both are more specific test cases can be built if needed. I don't wish to replicate the same changes across PRs in regards to fixing the pr and issue fetching etc

Refactors:
- direct config access no need for `timers` and `misc` props
- `isTaskStale` uses `ms` same as config items and more readable

with config: 
```yml
      - plugin: http://localhost:4000
        runsOn: ["issue_comment.created"]
        with:
          reviewDelayTolerance: "3 Days"
          taskStaleTimeoutDuration: "15 Days"
          maxConcurrentTasks: 3
          startRequiresWallet: false
```

QA:

- [More than 15 days old](https://github.com/ubq-testing/bot-ai/issues/50)
- [Less than 15 days old](https://github.com/ubq-testing/bot-ai/issues/51)

PR #24, opened 2024-08-20
feat: notify self assign

### LINKED ISSUE #12: Notify user on the deadline when assigned manually
V1 will notice a user of its deadline when it is assigned through the UI using and bypassing the `/start` command, see here: https://github.com/ubiquibot/conversation-rewards/issues/4#issuecomment-2237802296

V2 should also be notifying the user for the deadline on self-assign, or generally UI assign.

Also applies on UI assign, example:  https://github.com/ubiquibot/conversation-rewards/issues/81#issuecomment-2290677866

gentlementlegen:
Resolves #12 

Depends on #21, #19

QA: https://github.com/Meniole/command-start-stop/issues/1#issuecomment-2297967853


src/handlers/proxy.ts
0x4007, line 1: What is this file?

src/handlers/shared/start.ts
```diff
@@ -21,6 +22,14 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     throw new Error("Issue is a parent issue");
   }
 
+  const hasBeenPreviouslyUnassigned = await hasUserBeenUnassigned(context);
+
+  if (hasBeenPreviouslyUnassigned) {
+    const log = logger.error("You were previously unassigned from this task. You cannot reassign yourself.", { sender });
```
0x4007, line 28:
```suggestion
    const log = logger.error("You were previously unassigned from this task. You can not reassign yourself.", { sender });
```

src/handlers/shared/start.ts
```diff
@@ -109,5 +123,5 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
     ].join("\n") as string
   );
 
-  return { output: "Task assigned successfully" };
+  return { content: "Task assigned successfully", status: "ok" };
```
0x4007, line 126:
```suggestion
  return { content: "Task assigned successfully", statusCode: 200 };
```

HTTP status codes are far more concise and precise. Consider using these?

src/handlers/shared/stop.ts
```diff
@@ -13,7 +14,7 @@ export async function stop(context: Context, issue: Context["payload"]["issue"],
   if (!userToUnassign) {
     const log = logger.error("You are not assigned to this task", { issueNumber, user: sender.login });
     await addCommentToIssue(context, log?.logMessage.diff as string);
-    return { output: "You are not assigned to this task" };
+    return { content: "You are not assigned to this task", status: "ok" };
```
0x4007, line 17:
```suggestion
    return { content: "You are not assigned to this task", statusCode: 200 };
```

src/handlers/shared/stop.ts
```diff
@@ -39,5 +40,5 @@ export async function stop(context: Context, issue: Context["payload"]["issue"],
   });
 
   await addCommentToIssue(context, unassignedLog?.logMessage.diff as string);
-  return { output: "Task unassigned successfully" };
+  return { content: "Task unassigned successfully", status: "ok" };
```
0x4007, line 43:
```suggestion
  return { content: "Task unassigned successfully", statusCode: 200 };
```

src/types/env.ts
```diff
@@ -4,6 +4,7 @@ import { StaticDecode } from "@sinclair/typebox";
 export const envSchema = T.Object({
   SUPABASE_URL: T.String(),
   SUPABASE_KEY: T.String(),
+  APP_ID: T.String(),
```
0x4007, line 7:
```suggestion
  APP_ID: T.Number(),
```

These are numbers why not encode them as such?

src/handlers/shared/stop.ts
0x4007: Although this doesn't feel like `200` would be the correct "http status code" but you get the idea.

src/handlers/shared/stop.ts
0x4007: Same here I feel like there might be a more accurate status code to use.

src/types/env.ts
gentlementlegen: You should comment on these to the related PR, I did not authored these.
https://github.com/ubiquibot/command-start-stop/pull/21

src/handlers/shared/start.ts
gentlementlegen: It is arbitrary and not related to HTTP requests, this is just info for the Kernel, but sure can change it.

src/handlers/proxy.ts
gentlementlegen: It contains proxy calls for events.

0x4007: What is a "proxy call"?

src/handlers/shared/start.ts
0x4007: It makes sense to follow that standard.

src/types/env.ts
0x4007: I am confused because its not included in that pull. Besides, your pull indicates the change, or else it wouldn't appear on this diff.

src/handlers/proxy.ts
gentlementlegen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy 😄

Keyrxng: I would appreciate comments in this file and I think newer contributors would too. The first time I saw it in another plugin I thought "this looks like heavy artillery"

0x4007: Yeah I just had a quick tutoring session with ChatGPT for this.

Keyrxng: Seems to me the same could be achieved with a `if` or `switch` case but that may be a naïve opinion

src/types/env.ts
gentlementlegen: That's why I have it as a draft, and in the description I put "Depends on #21" which has to be merged and this one rebased. If it is not included then I forked from the wrong branch.

src/handlers/proxy.ts
```diff
@@ -0,0 +1,25 @@
+import { Context, Env, SupportedEventsU } from "../types";
+import { userSelfAssign, userStartStop } from "./user-start-stop";
+
+export interface Result {
+  status: "ok" | "failed" | "skipped";
```
0x4007, line 5: If there are only three statuses maybe the HTTP status code idea is overkill. What do you think makes sense @gentlementlegen ? I assumed this was some type of plugin standard.

src/handlers/proxy.ts
0x4007: > 	•	get is a trap for the get operation (i.e., property access). When you access a property on proxy, the trap intercepts the operation.

https://chatgpt.com/share/915e1875-a26a-4deb-b9fc-705e59f0212a

Keyrxng: > > •	get is a trap for the get operation (i.e., property access). When you access a property on proxy, the trap intercepts the operation.
> 
> https://chatgpt.com/share/915e1875-a26a-4deb-b9fc-705e59f0212a

But all this proxy does is convolute what could be a simple `if` or `switch` statement as we do not perform any mutation or action within the proxy other than calling one of two functions. `rpc-handler` uses a `Proxy` but it does mutate and control logic so it makes sense there, here it seems very OP.

I guess my point is, are we adopting this approach across plugins rather than the previous method of using an `if` for three `SupportedEvents` or less and a `switch` for three or more?

gentlementlegen: Sure a switch can work. Having the proxy enforces implementing all the supported events that would be the biggest difference.

src/handlers/proxy.ts
gentlementlegen: It's not standardized as these types should actually come from the Kernel. I don't remember if we finally exported these, I will have a look.

src/handlers/proxy.ts
Keyrxng: I am not against the Proxy but I think if we are adopting it as standard we should comment it clearly so that no-one else has to ask GPT what's going here

> Having the proxy enforces implementing all the supported events that would be the biggest difference.

This is a good reason for it.

---

I cannot shake the thought: if events are arbitrary and not defined in our `SupportedEvents` the `switch` and the `Proxy` will not run. 

If plugins are built to work on any event like `user-activity-watcher` I think you said @gentlementlegen then the proxy would need to handle all events. 

Plugins which cannot run on any event need to be restricted to only their supported events at the plugin config level I think.

gentlementlegen:
New QA: https://github.com/Meniole/command-start-stop/issues/10#issuecomment-2323796685

src/handlers/proxy.ts
gentlementlegen: I can make it a switch case, I just think that the more actions we should handle the more the switch case will be massive so that was to make code lighter. The proxy is needed because eventually the called key might not be in the object. In OOP it would just be a pointer to function object, which would be less fancy.

src/handlers/user-start-stop.ts
```diff
@@ -11,13 +17,22 @@ export async function userStartStop(context: Context): Promise<{ output: string
...
+}
+
+export async function userSelfAssign(context: Context): Promise<Result> {
+  const { payload } = context;
+  const { issue } = payload;
+  const deadline = getDeadline(issue);
+
+  const users = issue.assignees.map((user) => `@${user?.login}`).join(", ");
+
+  await addCommentToIssue(context, `${users} the deadline is at ${deadline}`);
```
whilefoo, line 36: won't `getDeadline` return null if price label is not present?

whilefoo (changes requested):

src/handlers/user-start-stop.ts
gentlementlegen: Good catch, I guess I'll totally skip if no deadline is set.

whilefoo (approved):

src/handlers/shared/start.ts
```diff
@@ -75,17 +75,17 @@ export async function start(context: Context, issue: Context["payload"]["issue"]
   }
 
   // get labels
-  const labels = issue.labels;
+  const labels = issue.labels ?? [];
   const priceLabel = labels.find((label: Label) => label.name.startsWith("Price: "));
 
   if (!priceLabel) {
     throw new Error(logger.error("No price label is set to calculate the duration", { issueNumber: issue.number }).logMessage.raw);
```
Keyrxng, line 82: For some reason this PR is not posting a comment but my PR I just opened fresh from `development` does post a comment no problem

https://github.com/ubq-testing/command-start-stop/issues/19

Keyrxng:
Unsure exactly what's causing it considering the bubble-up logic remains unchanged as far as I can tell other than the addition of the proxy but it looks like it's throw the error instead of passing it to our comment wrapper

Mine: 
image
Yours:
image

Keyrxng (changes requested):
There seems to be an issue with posting error comments

https://github.com/ubiquibot/command-start-stop/issues/34#issuecomment-2326108375

gentlementlegen:
> There seems to be an issue with posting error comments
> 
> [#34 (comment)](https://github.com/ubiquibot/command-start-stop/issues/34#issuecomment-2326108375)

Is it directly related to this pull-request or not? If so I will fix here, if not I'll do that on another PR.

Keyrxng:
for a `Time: <1 Hour` task we seem to be getting a deadline of `now()` 

###### I swear I don't use Microsoft Copilot
image

and `Time: <2 hour` gets a time of `now() + 1` 

image

Is this intended? 

---

I guess this is a separate issue below.

But didn't we used to give a buffer as opposed to using the raw time string? I bring this up because if contributors start to get negative XP for being beyond the deadline or lesser rewards because of it then surely the opposite should be true for tasks which get "completed" within the deadline. 

"Completed" is also subjective here, is that time til PR or time til closed-as-complete?

gentlementlegen:
@Keyrxng time is given in UTC, so I don't know the time zone you are in but that should be the reason. If not, I didn't change the code logic (I think) to get the deadline: https://github.com/ubiquibot/command-start-stop/pull/24/files#diff-0d8c330f3d4a9c9fe27deaf9eb73ede2d9bb6016940d6d24cd112688d819ecf5L15

And I think completed should be PR opening + ready to review, being merged in 1 hours sounds unrealistic.

Keyrxng:
> I didn't change the code logic (I think) to get the deadline

That'll be my mistake then my bad. My timezone is UTC+00 which I'm sure is the same as the reminder, do you think you could add a `+ 1` so that it's achievable to complete a 1hr task within the deadline?

> And I think completed should be PR opening + ready to review, being merged in 1 hours sounds unrealistic.

Agreed lol. And with issue/PR activity bumping it forward technically it's not a strict deadline either so calculating a bonus/negative based on the deadline seems like it will be a difficult task actually.

gentlementlegen:
@Keyrxng sure I can check that time issue and add some tests within a separate pull request.

Probably should use some averages / lean methods for xp calculations.

0x4007:
What is this about timing and XP? There is no approved proposal related to this. 

XP is derived only from dollar rewards, plus adjustments. 

No dollar rewards are related to deadlines and timing.

PR #29, opened 2024-08-28
feat: added worker deploy / delete capabilities

gentlementlegen:
Relates to https://github.com/ubiquibot/plugins-wishlist/issues/36


whilefoo (approved):
this will change the URL of the production worker right?

gentlementlegen:
@whilefoo It would append `-main` to it yes, I will fix the configurations afterwards. This won't get effective right away since these changes should be introduced in our Worker Plugins.

PR #33, opened 2024-08-30
fix: fixed /stop formatting

### LINKED ISSUE #32: Text formatting issue on `/stop` if the user is not assigned
There seem to be an issue with the displayed text in this scenario:
https://github.com/ubiquibot/sandbox/issues/21#issuecomment-2315946988

The output should be
```diff
! You are not assigned to this task
 ```

### PR DESCRIPTION
**Pull request by cohow**
Resolves #32

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -105,7 +105,7 @@ describe("User start/stop", () => {
 
     context.adapters = createAdapters(getSupabase(), context);
 
-    await expect(userStartStop(context as unknown as Context)).rejects.toThrow("```diff\n! You are not assigned to this task\n```");
+    await expect(userStartStop(context as unknown as Context)).rejects.toThrow("You are not assigned to this task");
   });
 
   test("User can't stop an issue without assignees", async () => {
@@ -115,7 +115,7 @@ describe("User start/stop", () => {
     const context = createContext(issue, sender, "/stop");
     context.adapters = createAdapters(getSupabase(), context as unknown as Context);
 
-    await expect(userStartStop(context as unknown as Context)).rejects.toThrow("```diff\n! You are not assigned to this task\n```");
+    await expect(userStartStop(context as unknown as Context)).rejects.toThrow("You are not assigned to this task");
   });
 
   test("User can't start an issue that's already assigned", async () => {
--- a/src/handlers/shared/stop.ts
+++ b/src/handlers/shared/stop.ts
@@ -11,7 +11,7 @@ export async function stop(context: Context, issue: Context["payload"]["issue"],
   const userToUnassign = assignees.find((assignee: Partial<Assignee>) => assignee?.login?.toLowerCase() === sender.login.toLowerCase());
 
   if (!userToUnassign) {
-    throw new Error(logger.error("You are not assigned to this task", { issueNumber, user: sender.login })?.logMessage.diff as string);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by Keyrxng — CHANGES_REQUESTED** [review]
Thanks for the PR @cohow 

This doesn't affect the working logic that's causing the issue

https://github.com/ubiquibot/command-start-stop/blob/a094a8c683aee3b2898049fef39f429d5e312a5d/src/handlers/shared/stop.ts#L14

`})?.logMessage.diff as string)` 

the problem is that the error is throwing the diff string which is bubbling up into the catch-all and then it's also being wrapped in a code block.

To fix the issue use `logMessage.raw` instead, only `addCommentToIssue()` should use `.diff` any errors should use `.raw`.

**Comment by cohow** [comment]
should be good to go now I believe

> To fix the issue use `logMessage.raw` instead, only `addCommentToIssue()` should use `.diff` any errors should use `.raw`.

I'll make sure to remember that for any future issue.

Thanks!

**Comment by gentlementlegen** [comment]
Seems to work fine when tested on my organization: https://github.com/Meniole/command-start-stop/issues/10#issuecomment-2323142275

@cohow next time please provide your own tests as well.

**Review by gentlementlegen — APPROVED** [review]

**Review by gentlementlegen — CHANGES_REQUESTED** [review]
Will approve when commit is renamed.

File: tests/main.test.ts [hunk 1]
```diff
@@ -565,7 +565,7 @@ function createContext(
       reviewDelayTolerance: "3 Days",
       taskStaleTimeoutDuration: "30 Days",
       maxConcurrentTasks: 3,
-      startRequiresWallet,
+      startRequiresWallet: false,
```
gentlementlegen, line 568: Why was this changed?

cohow: while rebasing it seems like i merged a new commit from development branch that includes that change, i’ll revert and commit new changes soon.

cohow:
@gentlementlegen requested changes have been done, sorry about the delay I fell asleep yesterday haha.

Keyrxng:
> gentlementlegen requested changes have been done, sorry about the delay I fell asleep yesterday haha.

@cohow take a look at [this cheatsheet](https://gist.github.com/Zekfad/f51cb06ac76e2457f11c80ed705c95a3) for conventional commits. Using `--no-verify` when committing will bypass our checks which would highlight that your commit naming is wrong. The logic is all good though

cohow:
> > gentlementlegen requested changes have been done, sorry about the delay I fell asleep yesterday haha.
> 
> @cohow take a look at [this cheatsheet](https://gist.github.com/Zekfad/f51cb06ac76e2457f11c80ed705c95a3) for conventional commits. Using `--no-verify` when committing will bypass our checks which would highlight that your commit naming is wrong. The logic is all good though

Okay I think I finally got it right this time... I forgot to add fix: before my apologies. 

Interestingly though other repos like https://github.com/ubiquibot/conversation-rewards would not let me commit without a proper commit name. Would be very nice to implement across all repos.

gentlementlegen:
Supposedly the webhooks should run with husky before you commit. Maybe there is something wrong with the setup on this repo, thank you for fixing.

gentlementlegen (approved):

Keyrxng:
> Would be very nice to implement across all repos.

I actually tested your commit title locally in this repo without using `--no-verify` and it complained and didn't let me commit so I'm certain it is implemented across all repos as it comes from our `ts-template` which basically every repo is forked from in one way or another, v strange lmk if you notice it again ty

PR #44, opened 2024-09-09
fix: added test for start with carriage returns

gentlementlegen:
Resolves https://github.com/ubiquity/ubiquibot-kernel/issues/105
QA: https://github.com/Meniole/command-start-stop/issues/16

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -89,7 +89,7 @@ describe("User start/stop", () => {
 
     const { content } = await userStartStop(context);
 
-    expect(content).toEqual(SUCCESS_MESSAGE);
+    expect(content).toEqual("Task unassigned successfully");
   });
 
   test("Stopping an issue should close the author's linked PR", async () => {
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -72,7 +72,7 @@ describe("User start/stop", () => {
 
     const { content } = await userStartStop(context);
 
-    expect(content).toEqual("Task assigned successfully");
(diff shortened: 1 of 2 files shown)


gentlementlegen:
I merge this as it is related to https://github.com/ubiquibot/command-start-stop/pull/44. Let me know if there is any fix needed I'll take care of it.

tests/main.test.ts
```diff
@@ -44,7 +46,20 @@ describe("User start/stop", () => {
 
     const { content } = await userStartStop(context);
 
-    expect(content).toEqual("Task assigned successfully");
+    expect(content).toEqual(SUCCESS_MESSAGE);
+  });
+
+  test("User can start an issue with trimmed command", async () => {
+    const issue = db.issue.findFirst({ where: { id: { equals: 1 } } }) as unknown as Issue;
+    const sender = db.users.findFirst({ where: { id: { equals: 1 } } }) as unknown as PayloadSender;
```
0x4007, line 54: This uses a database? It should be querying the GitHub api only.

gentlementlegen: This is a db used for faking API calls: https://github.com/ubiquibot/command-start-stop/blob/19743ea67faaaea32937791787f000701b8d1cb5/tests/main.test.ts#L3

0x4007: That doesn't seem to make sense to me. We should be using a mock http server not a database.

0x4007: No idea about all this

PR #50, opened 2024-09-14
chore: new assign design

0x4007:
From my fork to ubiquibot


0x4007:
This deployment isn't visible on the dashboard I don't understand

https://github.com/ubiquibot/command-start-stop/actions/runs/10862228060/job/30148166581

0x4007:
The only difference is that I recreated `yarn.lock` and now it says `BOT_USER_ID` is not set. Its set on both `ubiquibot` and `0x4007` repos...

0x4007:
@gentlementlegen I'm running out of time but the deploys I can't get to work.

PR #52, opened 2024-09-16
fix: schema decode defaults

### LINKED ISSUE #51: Fix default schema values
It appears to me that the logic to decode the schema is wrong, and this would be introduced by Jordan's PR addition regarding the configuration accepting the amount of parallel tasks allowed for each user. If not provided, the default configuration fails which is why it is failing in your last try, on ubiquity / ubiquibot orgs, but not on mine where I provide the full config. Working on a fix right now.

_Originally posted by @gentlementlegen in https://github.com/ubiquibot/permit-generation/issues/60#issuecomment-2351912070_
            

The default values when not provided via the config are:
```
{ admin: Infinity, member: 10, contributor: 2 }
```
`Infinity` does not satisfy `T.Record(T.String(), T.Integer())` since it is not an integer value.

Tasks to be carried out:
- properly log the errors so it is easier to identify
- add a test related to the default configuration if none
- fix the default values

gentlementlegen:
Resolves #51

### DIFF
--- a/src/types/plugin-input.ts
+++ b/src/types/plugin-input.ts
@@ -1,6 +1,6 @@
-import { SupportedEvents, SupportedEventsU } from "./context";
 import { StaticDecode, Type as T } from "@sinclair/typebox";
 import { StandardValidator } from "typebox-validators";
+import { SupportedEvents, SupportedEventsU } from "./context";
 
 export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU extends SupportedEvents[T] = SupportedEvents[T]> {
   stateId: string;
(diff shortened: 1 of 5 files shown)


src/types/plugin-input.ts
```diff
@@ -11,12 +11,24 @@ export interface PluginInputs<T extends SupportedEventsU = SupportedEventsU, TU
   ref: string;
 }
 
+function maxConcurrentTasks() {
+  return T.Transform(T.Record(T.String(), T.Integer(), { default: { member: 10, contributor: 2 } }))
+    .Decode((value) => {
+      // If admin is omitted, defaults to infinity
```
0x4007, line 17: Just remove everything admin related. Simplifies the code and it was out of scope in the first place.

0x4007 (changes requested):

src/types/plugin-input.ts
gentlementlegen: Possible roles are "admin", "member", "collaborator" + any other custom role added to the org. I can't remove admin logic, I could just hardcode a condition "if admin then infinity" which wouldn't change the code much I believe, because the admin case has to be handled.

Keyrxng (approved):
Works as intended. Infinity as default and configurable.

image
image

whilefoo (approved):

PR #57, opened 2024-09-23
feat: pull edit handler

### LINKED ISSUE #55: Verify that the user gets assigned on PR open
@Keyrxng rfc

What I think happened is:
- no user was assigned
- a pull-request was linked to this issue, which triggers an assignment of the user opening that PR
- the user somehow didn't get assigned (you can see from the timeline only the start message was triggered)
- `/start` assigned the user and displayed the message again

It should have just said "You are already assigned to this task" because it should have been the case.

_Originally posted by @gentlementlegen in https://github.com/ubiquibot/assistive-pricing/issues/32#issuecomment-2362687081_

gentlementlegen:
Resolves #55

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Start | Stop",
   "description": "Assign or un-assign yourself from an issue.",
-  "ubiquity:listeners": ["issue_comment.created", "issues.assigned", "issues.unassigned", "pull_request.opened"],
+  "ubiquity:listeners": ["issue_comment.created", "issues.assigned", "issues.unassigned", "pull_request.opened", "pull_request.edited"],
   "commands": {
     "start": {
       "ubiquity:example": "/start",
--- a/src/handlers/user-start-stop.ts
+++ b/src/handlers/user-start-stop.ts
@@ -44,7 +44,7 @@ export async function userSelfAssign(context: Context<"issues.assigned">): Promi
   return { status: HttpStatusCode.OK };
 }
 
-export async function userPullRequest(context: Context<"pull_request.opened">): Promise<Result> {
+export async function userPullRequest(context: Context<"pull_request.opened" | "pull_request.edited">): Promise<Result> {
   const { payload } = context;
   const { pull_request } = payload;
   const { owner, repo } = getOwnerRepoFromHtmlUrl(pull_request.html_url);
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -31,6 +31,8 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
       case "issues.assigned":
         return await userSelfAssign(context as Context<"issues.assigned">);
       case "pull_request.opened":
+        return await userPullRequest(context as Context<"pull_request.edited">);
+      case "pull_request.edited":
         return await userPullRequest(context as Context<"pull_request.opened">);
       case "issues.unassigned":
         return await userUnassigned(context);
--- a/src/types/context.ts
+++ b/src/types/context.ts
@@ -6,7 +6,7 @@ import { createAdapters } from "../adapters";
 import { Env } from "./env";
 import { Logs } from "@ubiquity-dao/ubiquibot-logger";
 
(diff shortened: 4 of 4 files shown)


manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Start | Stop",
   "description": "Assign or un-assign yourself from an issue.",
-  "ubiquity:listeners": ["issue_comment.created", "issues.assigned", "issues.unassigned", "pull_request.opened"],
+  "ubiquity:listeners": ["issue_comment.created", "issues.assigned", "issues.unassigned", "pull_request.opened", "pull_request.edited"],
```
0x4007, line 4: This is for adding `resolves #xxx` ?

0x4007 (commented):
Don't understand this change

0x4007:
Your update commentary needs to be under the pull not the issue 

> They all worked fine. The only problem I noticed and when you open a PR, then change the description and link an issue, the bot won't assign and post the comment, which believe should be fixed (dunno if v1 handled that).

So you're saying now the bot reads the issue number from the closing keyword and automatically assigns? I don't understand what the point of adding pull body edit is

gentlementlegen:
It fixes the scenario where you open a PR (draft or not) with an empty description, and then you edit it adding the `Resolves #xx`) . Within github, it will link your PR as closing of the issue, but if we do not observe the event on the plugin, the kernel won't assign the user to the issue.

0x4007:
It should only be on the linked event. Right? It will "link" when they fix the pull body

gentlementlegen:
You can try it, open a pull-request with an empty body, then modify it to link it to a PR, GitHub will properly create a link between the pull-request and the issue. There is no linked event afaik, so checking the pull-request modified allows us to check for the linked issue, if any.

0x4007 (approved):

gentlementlegen:
@0x4007 Seems that Jest is not ok with the changes? Was it broken before you closed the PR? I'll fix this.

Keyrxng:
I just QA'd development and it is possible to have more than one linked issue in the PR and in my case it assigned me to both tasks and I thought of legit pieces of text that may use any of the below prefixes. I think we should handle this case but I can't think of a guaranteed way to do it.

image

image

I think the best way to do it maybe is to use the reference which appears first in the PR body? Otherwise if all the conditions of `/start` are valid then the user would be assigned to all the tasks linked which are.

0x4007:
Can you link your qa I don't understand your explanation.

gentlementlegen:
@Keyrxng why not using GraphQl? It handles multiple linked pull-requests.

Keyrxng:
- https://github.com/ubq-testing/command-start-stop/pull/23

I first comment with only "Resolves ..." and then I edit it to add more context regarding a related issue to which I get [assigned](https://github.com/ubq-testing/command-start-stop/issues/22). Because the conditions of `/start` were valid I was assigned the task while my intention may not have been that. 

It makes it so that you must be aware of not using these prefixes otherwise you might end up assigned.

> why not using GraphQl? It handles multiple linked pull-requests.

This is using the `development` branch which is using GraphQL. I'm not sure how you'd be able to accurately deduce from the GraphQL issue response objects which referenced task should be assigned as both might exist in the same repo etc. Unless you can tell from the GraphQL fetch which appears first ofc.

0x4007:
> It makes it so that you must be aware of not using these prefixes otherwise you might end up assigned.

Then don't use the prefixes? Or just unassign yourself if its a problem?

Keyrxng:
> Then don't use the closing keywords? Or just unassign yourself if its a problem?

My point was that contributors partners etc will all need to be aware of these little gotchas or it's going to happen to others out in the wild, obviously I know to avoid it so that it doesn't happen to me.

But if it's a non-issue then no worries just thought I'd shine a light on it

0x4007:
If it follows GitHub behavior it's fine

I think if the pull body starts with a closing keyword then GitHub will link it

I'm not sure about in the middle but even so it's so specific that it's a non issue. It's not gonna happen on accident. 

Just test and match GitHub behavior

PR #58, opened 2024-09-23
chore: fixed tests

gentlementlegen:
A type was wrongly deduced, added a check so TS is happy.

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -1,15 +1,16 @@
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect } from "@jest/globals";
 import { drop } from "@mswjs/data";
+import { createClient } from "@supabase/supabase-js";
+import { cleanLogString, Logs } from "@ubiquity-dao/ubiquibot-logger";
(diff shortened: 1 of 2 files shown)


tests/main.test.ts
```diff
@@ -599,7 +600,7 @@ const maxConcurrentDefaults = {
 
 function createContext(
   issue: Record<string, unknown>,
-  sender: Record<string, unknown>,
+  sender: Record<string, unknown> | undefined,
```
0x4007, line 603: Undefined seems impossible?

gentlementlegen: It seems that it can theoretically happen on a `pull_request.modified` payload? Maybe if it is modified by some automation.

0x4007: Seems wrong doesn't make sense to me

gentlementlegen: That's how GitHub code is written, not much I can do:
Screenshot 2024-09-24 at 08 03 44

0x4007 (approved):

PR #65, opened 2024-10-10
feat: task-restriction

### LINKED ISSUE #62: Collaborator Gating Based On Label
- Since we still are ways off to implementing XP, we should implement a simpler gating system. 
- Similar to how we manage max concurrent tasks per role, we should also have a config to gate based on the label. 
- There's no point in making a config for admin.

```yml
  - uses:
    - plugin: https://ubiquibot-assistive-pricing-development.ubiquity.workers.dev
      with:
        labels:
          time:
            - "Time: <1 Hour"
              collaboratorOnly: false
            - "Time: <2 Hours"
              collaboratorOnly: false
            - "Time: <4 Hours"
              collaboratorOnly: true
            - "Time: <1 Day"
              collaboratorOnly: true
            - "Time: <1 Week"
              collaboratorOnly: true
            - "Time: <1 Month"
              collaboratorOnly: true
          priority:
            - "Priority: 1 (Normal)"
              collaboratorOnly: false
            - "Priority: 2 (Medium)"
              collaboratorOnly: false
            - "Priority: 3 (High)"
              collaboratorOnly: true
            - "Priority: 4 (Urgent)"
              collaboratorOnly: true
            - "Priority: 5 (Emergency)"
              collaboratorOnly: true
        basePriceMultiplier: 2
        publicAccessControl:
          setLabel: true
          fundExternalClosedIssue: false
```

We could even embed metadata in the label description, although for this purpose maybe via the config here is simple enough.

gentlementlegen:
Resolves #62

### DIFF
--- a/src/handlers/shared/start.ts
+++ b/src/handlers/shared/start.ts
@@ -1,4 +1,5 @@
 import { Context, ISSUE_TYPE, Label } from "../../types";
+import { isUserCollaborator } from "../../utils/get-user-association";
 import { addAssignees, addCommentToIssue, getAssignedIssues, getAvailableOpenedPullRequests, getTimeValue, isParentIssue } from "../../utils/issue";
 import { HttpStatusCode, Result } from "../result-types";
 import { hasUserBeenUnassigned } from "./check-assignments";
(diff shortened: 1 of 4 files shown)


src/handlers/shared/start.ts
```diff
@@ -82,14 +83,26 @@ export async function start(
...
   if (!priceLabel) {
     throw logger.error("No price label is set to calculate the duration", { issueNumber: issue.number });
   }
 
+  // Checks if non-collaborators can be assigned to the issue
+  for (const label of labels) {
+    if (label.description?.includes("collaborator only")) {
+      for (const user of toAssign) {
+        if (!(await isUserCollaborator(context, user))) {
+          throw logger.error("Non-collaborators cannot be assigned to this issue", {
```
0x4007, line 98:
```suggestion
          throw logger.error("Only collaborators can be assigned to this issue", {
```

0x4007:
Just make sure that you don't change the Price label descriptions with the collaborator only warning.

gentlementlegen:
@0x4007 should not happen
https://github.com/ubiquity-os-marketplace/daemon-pricing/blob/7918d516322c72b8ec31a49851aec28ce6686b0e/src/handlers/sync-labels-to-config.ts#L77

PR #35, opened 2024-09-03
Use something other than APP_ID for botId

Keyrxng:
Resolves #34 

- `APP_ID` cannot be used to ID the bot
- remove two comment postings
- pass `decodedEnv`

PR #66, opened 2024-10-20
feat: task limit improvement

### LINKED ISSUE #64: Task Limit Improvements
This needs to:

1. Change to warning syntax vs diff syntax
2. Link all the issues I'm currently assigned to in a comment so I can unassign myself or close them out. 
3. Only check assigned issues within the network, but this should be configurable to either be just within the repo, org, or network

---

```diff
! You have reached your max task limit. Please close out some tasks before assigning new ones.
```

_Originally posted by @ubiquity-os[bot] in https://github.com/ubiquity/work.ubq.fi/issues/120#issuecomment-2404264757_
            

>[!WARNING]
> You have reached your max task limit. Please close out some tasks before assigning new ones.

ariesgun:
Resolves #64


src/utils/issue.ts
```diff
@@ -15,7 +15,7 @@ export async function getAssignedIssues(context: Context, username: string): Pro
   try {
     return await context.octokit
       .paginate(context.octokit.search.issuesAndPullRequests, {
-        q: `org:${payload.repository.owner.login} assignee:${username} is:open is:issue`,
+        q: `org:ubiquity org:ubiquity-os-marketplace org:ubiquity-os assignee:${username} is:open is:issue`,
```
0x4007, line 18: I think the most robust way to get all "within network" is to get the list of orgs that installed the bot

0x4007 (changes requested):

0x4007:
You also should post QA

ariesgun:
> You also should post QA

sorry, new here. what do you mean by "post QA"?

0x4007:
Prove that it works on your org

src/utils/issue.ts
```diff
@@ -173,7 +179,7 @@ export async function addAssignees(context: Context, issueNo: number, assignees:
 export async function getAllPullRequests(context: Context, state: "open" | "closed" | "all" = "open", username: string) {
   const { payload } = context;
   const query: RestEndpointMethodTypes["search"]["issuesAndPullRequests"]["parameters"] = {
-    q: `org:${payload.repository.owner.login} author:${username} state:${state}`,
+    q: `org:ubiquity org:ubiquity-os-marketplace org:ubiquity-os author:${username} state:${state}`,
```
0x4007, line 182: Here too

gentlementlegen: Hardcoding these means third parties won't be able to use this plugin.

ariesgun: Yeah, working on this one... 
Trying to test it in a custom org, but for some reasons it cant get the max task limit.. checking it

ariesgun:
Test Result:

https://github.com/ubiquity-ariesgun/command-start-stop/issues/9#issuecomment-2427166082

src/utils/issue.ts
ariesgun: Addressed

src/utils/issue.ts
ariesgun: Addressed

ariesgun:
> Prove that it works on your org. This would be a link to a github comment

Test Result:

https://github.com/ubiquity-ariesgun/command-start-stop/issues/9#issuecomment-2427166082

src/plugin.ts
```diff
@@ -6,17 +6,27 @@ import { createAdapters } from "./adapters";
...
 import { Context, Env, PluginInputs } from "./types";
 import { addCommentToIssue } from "./utils/issue";
+import { createAppAuth } from "@octokit/auth-app";
 
 export async function startStopTask(inputs: PluginInputs, env: Env) {
   const customOctokit = Octokit.plugin(paginateGraphQL);
   const octokit = new customOctokit({ auth: inputs.authToken });
   const supabase = createClient(env.SUPABASE_URL, env.SUPABASE_KEY);
 
+  const jwtOctokit = new Octokit({
```
gentlementlegen, line 16: I don't think this is correct. Why would we need two Octokit instances?

ariesgun: hmm indeed.. let me check if i can combine them.
To be able to use `listInstallations`, it needs JWT.

ariesgun: I cannot use the provided `inputs.authToken`, I don't think it is a JWT token. 

```
{
            "err": {
              "name": "HttpError",
              "status": 401,
              "response": {
                "url": "https://api.github.com/app/installations",
                "status": 401,
                "headers": {
                  "access-control-allow-origin": "*",
                  "access-control-expose-headers": "ETag, Link, Location, Retry-After, X-GitHub-OTP,
  X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource,
  X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type,
  X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset",
                  "content-length": "117",
                  "content-security-policy": "default-src 'none'",
                  "content-type": "application/json; charset=utf-8",
                  "date": "Mon, 21 Oct 2024 19:04:16 GMT",
                  "referrer-policy": "origin-when-cross-origin, strict-origin-when-cross-origin",
                  "server": "github.com",
                  "strict-transport-security": "max-age=31536000; includeSubdomains; preload",
                  "vary": "Accept-Encoding, Accept, X-Requested-With",
                  "x-content-type-options": "nosniff",
                  "x-frame-options": "deny",
                  "x-github-media-type": "github.v3; format=json",
                  "x-github-request-id": "C484:3BDE26:ADC1C3:AFD8B6:6716A5B0",
                  "x-xss-protection": "0"
                },
                "data": {
                  "message": "A JSON web token could not be decoded",
                  "documentation_url": "https://docs.github.com/rest",
                  "status": "401"
                }
              },
              "request": {
                "method": "GET",
                "url": "https://api.github.com/app/installations",
                "headers": {
                  "accept": "application/vnd.github.v3+json",
                  "user-agent": "octokit-rest.js/20.1.1 octokit-core.js/5.2.0 Cloudflare-Workers",
                  "authorization": "token [REDACTED]"
                },
                "request": {}
              }
            },
            "caller": "startStopTask"
          }
          
```
Not sure what the best way to solve this.

gentlementlegen: The octokit is logged in using a Github App token.

src/utils/issue.ts
Keyrxng: Afaik this isn't feasible @0x4007 because we need the kernel' private key which has been spoken about across a few issues but started [here](https://github.com/ubiquity-os-marketplace/command-start-stop/pull/35)

src/plugin.ts
Keyrxng: to list installs we need app private key do we not?

ariesgun: yes, we do need JWT to be exact.

0x4007: What we've done in the past is get it passed in from GitHub Actions. 

Get: https://github.com/ubiquity/cloudflare-deploy-action/blob/main/action.yml#L50-L56

Use: https://github.com/ubiquity/cloudflare-deploy-action/blob/main/action.yml#L62

0x4007:
> Test Result:
> 
> [ubiquity-ariesgun#9 (comment)](https://github.com/ubiquity-ariesgun/command-start-stop/issues/9#issuecomment-2427166082)

Currently looking into adjusting the display format:

---

> [!Warning]
> 
> You have reached your max task limit. Please close out some tasks before assigning new ones.


- ###### [ubiquity-os-marketplace/command-start-stop - Task Limit Improvements #64](https://www.github.com/ubiquity-os-marketplace/command-start-stop/issues/64)

```
- ###### [ubiquity-os-marketplace/command-start-stop - Task Limit Improvements #64](https://www.github.com/ubiquity-os-marketplace/command-start-stop/issues/64)
```

src/plugin.ts
Keyrxng: fair enough

ariesgun:
> > Test Result:
> > [ubiquity-ariesgun#9 (comment)](https://github.com/ubiquity-ariesgun/command-start-stop/issues/9#issuecomment-2427166082)
> 
> Currently looking into adjusting the display format:
> 
> Warning
> 
> You have reached your max task limit. Please close out some tasks before assigning new ones.
> 
> * ###### [ubiquity-os-marketplace/command-start-stop - Task Limit Improvements #64](https://www.github.com/ubiquity-os-marketplace/command-start-stop/issues/64)
> 
> ```
> - ###### [ubiquity-os-marketplace/command-start-stop - Task Limit Improvements #64](https://www.github.com/ubiquity-os-marketplace/command-start-stop/issues/64)
> ```

like this? https://github.com/ubiquity-ariesgun/command-start-stop/issues/9#issuecomment-2430213200

src/handlers/shared/start.ts
```diff
@@ -64,23 +64,49 @@ export async function start(
   teammates.push(sender.login);
 
   const toAssign = [];
+  let assignedIssues : any[]= [];
```
gentlementlegen, line 67: Do not use `any` and format properly.

src/plugin.ts
gentlementlegen: Still feels wrong to me that we need two octokit instances with different auth and private keys. @whilefoo rfc

src/handlers/shared/start.ts
```diff
@@ -64,23 +64,49 @@ export async function start(
...
-  if (error) {
-    throw logger.error(error, { issueNumber: issue.number });
+    await addCommentToIssue(context, `
+      
+> [!WARNING]
+> ${error}
+
+${issues}
+
+    `)
```
gentlementlegen, line 108:
```suggestion
`)
```

src/plugin.ts
Keyrxng: listing installs is for sure the most convenient and effective way to find out the entire partner network

hack 1: use `ubiquityos_app_slug` against the search API for comments and then filter a `Set` of the orgs which it has appeared in
hack 2: install a secondary app to avoid exposing kernel' pk (guaranteed it seems partner's will have to do this anyway)
hack 3: Auth as the kernel with `clientId` but I think that's the same as exposing the PK

ariesgun: With option 1, it means third parties won't be able to use the plugin right? 
Is it also an idea to query all issues in `devpool-directory` and built a set of orgs from the issues there?

Keyrxng: 1. partner's will use the same app slug, if you mean forks/development then they'd change to their own app slug.
2. That is also an effective suggestion

whilefoo: Do we need bot's token just to list all installations or also installation tokens from each installation?
One option would be an API endpoint on the kernel to list installations, I'd avoid using bot's private key other than in kernel if possible.

ariesgun: yes, we only need bot's token to list all installations
Regarding the API endpoint, is it something that is already possible on the kernel side? If so, I think it is better to move it there

src/utils/issue.ts
ariesgun: I would like to take a step back.
Can you provide some insights why we would like to search assigned issues within the network? I can't think of the reason why searching for issues within the same organization is not enough?
It won't be possible to list all orgs that installed the bot without using app private key.

Keyrxng: > Can you provide some insights why we would like to search assigned issues within the network?

To get a better picture of how over-assigned the contributor might be across multiple partners (eventually we'll have loads).

---

1. Global storage would enable us to track a userbase across all partners. The user object can be updated any time they are assigned an issue within the network and we can track with ease. This can serve us in hundreds of ways not just with this use-case.
2. Parse devpool json (better with all issues filled, theres a task for it) + search API for `username: pull_requests`.
3. Depending on the user' visibility settings you'll likely be able to do it with GraphQL API but I'm not as knowledgeable with that API, I use the official [GitHub GQL Explorer](https://docs.github.com/en/graphql/overview/explorer) to investigate when I need to.

ariesgun: I see, understood.

Based on your suggestion to use Github Search API, I played with it a bit and I think it is possible to search for issues where the bot has interacted with, using this query: `is:open is:issue commenter:ubiquity-os[bot] assignee:<user>`..
I think it can work, and we don't need to use the private key anymore. 

The question now is should I check for `ubiquity-os[bot]` explicitly or make it configurable in `.ubiquity-os.config.yml`.

Keyrxng: Whichever way you feel is most effective dude go for it.

https://github.com/ubiquity-os-marketplace/command-start-stop/blob/1ceebb363cf414b5a9ad00c101ea5e1d706424fb/src/types/env.ts#L9

Use this variable to find out the bot's name instead of a new config item but if not then I guess you'd need a new config item if using that approach but I feel the right thing to do is use it's ID vs a new config item.

ariesgun: Unfortunately cannot search for App with an `id`, it needs to be a `node_id`. And we don't have it,.

Keyrxng: > Unfortunately cannot search for App with an `id`, it needs to be a `node_id`. And we don't have it,.

ohh I was hoping that wasn't the case. You could use `author: <xx> is:pr` and do an org comparison that way instead otherwise @0x4007 or @gentlementlegen will approve the new config item.

ariesgun: With this approach, the idea is to compare with devpools.json right? Checking the json file, but it seems https://raw.githubusercontent.com/ubiquity/devpool-directory/refs/heads/development/devpool-issues.json is not available anymore?

And a question: shouldn't it be enough to just build the list of orgs using devpools.json? Is it useful to compare with the results from `author:<xxx> is:pr` query?

Keyrxng: https://github.com/ubiquity/devpool-directory/tree/__STORAGE__ - storage has it's own branch.

> And a question: shouldn't it be enough to just build the list of orgs using devpools.json? Is it useful to compare with the results from author:<xxx> is:pr query?

This fn is trying to find assigned issues so

- using search API to get all PRs, then filter, then get the linked issue via GQL or PR body parsing
- using search API with `assignee:keyrxng is:issue`, then just filter is another but it doesn't take PRs into consideration (not sure if that's something we should consider in this context)

option two is probs best.

whilefoo: > Use this variable to find out the bot's name instead of a new config item but if not then I guess you'd need a new config item if using that approach but I feel the right thing to do is use it's ID vs a new config item.

Actually the kernel/SDK exports the app id and user ID of the production bot. I was thinking that maybe this would be actually sent along to the plugin because it can change if you run your own kernel

> Pretty sure the REST api does not provide a method to search for a user with their user.id or node_id so you'd need to use the GQL API.

This is the api: https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user-using-their-id

ariesgun:
QA:

https://github.com/ubiquity-ariesgun/command-start-stop/issues/12#issuecomment-2439593830

tests/main.test.ts
```diff
@@ -634,6 +632,7 @@ export function createContext(
     },
     octokit: new octokit.Octokit(),
     eventName: "issue_comment.created" as SupportedEventsU,
+    organizations: ["ubiquity"],
```
0x4007, line 635: Seems suspiciously wrong. Should be dynamic based on installations.

src/plugin.ts
```diff
@@ -4,18 +4,36 @@ import { createClient } from "@supabase/supabase-js";
 import { LogReturn, Logs } from "@ubiquity-os/ubiquity-os-logger";
 import { createAdapters } from "./adapters";
 import { userPullRequest, userSelfAssign, userStartStop, userUnassigned } from "./handlers/user-start-stop";
-import { Context, Env, PluginInputs } from "./types";
+import { Context, Env, GitHubIssueSearch, PluginInputs } from "./types";
 import { addCommentToIssue } from "./utils/issue";
 
+async function listOrganizations(): Promise<string[]> {
+  const res = await fetch("https://raw.githubusercontent.com/ubiquity/devpool-directory/refs/heads/__STORAGE__/devpool-issues.json");
```
0x4007, line 11: I wish we had a mechanism to detect if this breaks to respond quickly because that system may be changed without notice in the future. I suppose for now you should throw a specific and descriptive error if it 404.

src/utils/issue.ts
0x4007: > Afaik this isn't feasible @0x4007 because we need the kernel' private key which has been spoken about across a few issues. It was ofc the telegram bot most recently [ubiquity-os-marketplace/ubiquity-os-kernel-telegram#2 (comment)](https://github.com/ubiquity-os-marketplace/ubiquity-os-kernel-telegram/pull/2#discussion_r1766516363)

Try this. Its available in the org secrets.

https://github.com/ubiquity/cloudflare-deploy-action/blob/3e07491a4785469af9eee58bf697203428c0f8ee/action.yml#L50-L56

ariesgun: Hmm.. So we use this

> > Afaik this isn't feasible @0x4007 because we need the kernel' private key which has been spoken about across a few issues. It was ofc the telegram bot most recently [ubiquity-os-marketplace/ubiquity-os-kernel-telegram#2 (comment)](https://github.com/ubiquity-os-marketplace/ubiquity-os-kernel-telegram/pull/2#discussion_r1766516363)
> 
> Try this. `APP_ID` and `APP_PRIVATE_KEY` have now been added to the @ubiquity-os-marketplace org secrets.
> 
> GET: https://github.com/ubiquity/cloudflare-deploy-action/blob/3e07491a4785469af9eee58bf697203428c0f8ee/action.yml#L50-L56
> 
> SET: https://github.com/ubiquity/cloudflare-deploy-action/blob/3e07491a4785469af9eee58bf697203428c0f8ee/action.yml#L62
> 
> Just make sure to get from secrets instead of it being "passed in" like it is here.

Hmm, I assume we can only access secrets in Github/ workflow. So, I imagine calling `listInstallations` inside a Github Workflow. I am not sure how to pass this information to the plugin..??

tests/main.test.ts
whilefoo: it's a unit test

0x4007:
Need to specifically handle 404

ariesgun:
> Need to specifically handle 404

Any specific action when is 404?
Right now it will log and write it as a comment in github in case of non-200

0x4007:
> I wish we had a mechanism to detect if this breaks to respond quickly because that system may be changed without notice in the future. I suppose for now you should throw a specific and descriptive error if it 404.

0x4007 (approved):

package.json
```diff
@@ -83,6 +83,5 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  },
-  "packageManager": "yarn@4.2.2"
```
gentlementlegen, line 87: @0x4007 Don't we want this? It can be `1.22` but setting it avoids collaborator to use the wrong package manager and encounter strange errors.

ariesgun: It was unintentional, reverted it

ubiquity-ubiquibot (changes requested):
I can't test this because there is no way to set the organizations you want to look for assigned items, and it seems to always target `ubiquity` related repos so it always let me assign to tasks without any limit. I set it to `1` in the configuration, yet I could start in two issues:
https://github.com/Meniole/command-start-stop/issues/39
https://github.com/Meniole/command-start-stop/issues/40

There is nothing in the configuration that let's me change the targeted repos, which doesn't seem to fulfill point three of the spec. Am I misunderstanding something? @0x4007 rfc

gentlementlegen:
I can't test this because there is no way to set the organizations you want to look for assigned items, and it seems to always target ubiquity related repos so it always let me assign to tasks without any limit. I set it to 1 in the configuration, yet I could start in two issues:
https://github.com/Meniole/command-start-stop/issues/39
https://github.com/Meniole/command-start-stop/issues/40

There is nothing in the configuration that let's me change the targeted repos, which doesn't seem to fulfill point three of the spec. Am I misunderstanding something? @0x4007 rfc

src/plugin.ts
```diff
@@ -24,6 +50,20 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
...
+  } catch (err) {
+    let errorMessage;
+    if (err instanceof LogReturn) {
+      errorMessage = err;
+      await addCommentToIssue(context, `${errorMessage?.logMessage.diff}\n<!--\n${sanitizeMetadata(errorMessage?.metadata)}\n-->`);
+    } else {
+      context.logger.error("An error occurred", { err });
+    }
+    return;
+  }
```
whilefoo, line 65: move this code to the try/catch block below to avoid duplicating catch block

whilefoo (changes requested):
forgot to submit review :D

src/handlers/shared/start.ts
```diff
@@ -64,23 +64,46 @@ export async function start(
...
+      `
+      
+> [!WARNING]
+> ${error}
+
+${issues}
+
+`
+    );
+    throw new Error(logger.error(error, { issueNumber: issue.number }).logMessage.raw);
```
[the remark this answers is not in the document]
ariesgun: So it stops executing the remaining code in this function.
Same behavior as the previous implementation except it throws an Error now so we won't see two comments on Github

src/plugin.ts
```diff
@@ -4,18 +4,36 @@ import { createClient } from "@supabase/supabase-js";
...
 import { addCommentToIssue } from "./utils/issue";
 
+async function listOrganizations(): Promise<string[]> {
+  const res = await fetch("https://raw.githubusercontent.com/ubiquity/devpool-directory/refs/heads/__STORAGE__/devpool-issues.json");
+  const devpoolIssues: GitHubIssueSearch["items"] = await res.json();
+  const orgsSet: Set<string> = new Set();
+
+  const urlPattern = /https:\/\/(github.com\/(\S+)\/(\S+)\/issues\/(\d+))/g;
+  devpoolIssues.forEach((issue) => {
+    const matches = [...issue.html_url.matchAll(urlPattern)][0];
```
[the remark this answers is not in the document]
ariesgun: indeed can be improved

0x4007:
Should be configurable 
Repo
Org
Network

Also that's wild that I can't delete or hide your review comment

gentlementlegen:
@0x4007 Why did you hide my comment (if that was you)?

0x4007:
Your review comment is the exact same content

gentlementlegen:
@0x4007 maybe you are confused because I first posted my comment with the wrong account and then hid it to repost with my own? I can't seem to find twice the same comment from myself.

0x4007:
image

gentlementlegen:
Yes that was with the wrong account, I hid that comment.

ariesgun:
> @ariesgun, this task has been idle for a while. Please provide an update.

ongoing

0x4007:
Can we merge @gentlementlegen @whilefoo

gentlementlegen:
@0x4007 The conflicts still need to be resolved first.

ariesgun:
Made the issues search configurable: repo, org, or network.
Fixed the merge conflict.

Network:
https://github.com/ubiquity-ariesgun/command-start-stop/issues/12#issuecomment-2453487597

Org:
https://github.com/ubiquity-ariesgun/command-start-stop/issues/12#issuecomment-2453486116

Repo:
https://github.com/ubiquity-ariesgun/command-start-stop/issues/12#issuecomment-2453485977

src/plugin.ts
```diff
@@ -16,6 +53,7 @@ export async function startStopTask(inputs: PluginInputs, env: Env) {
     eventName: inputs.eventName,
     payload: inputs.eventPayload,
     config: inputs.settings,
+    organizations: [],
```
gentlementlegen, line 56: This should not be here but in the configuration.

src/utils/issue.ts
```diff
@@ -11,13 +11,20 @@ export function isParentIssue(body: string) {
...
+      repoOrgQuery += `org:${org} `;
+    });
+  }
 
   try {
-    return await context.octokit
+    return context.octokit
       .paginate(context.octokit.rest.search.issuesAndPullRequests, {
-        q: `org:${payload.repository.owner.login} assignee:${username} is:open is:issue`,
+        q: `${repoOrgQuery} assignee:${username} is:open is:issue`,
```
gentlementlegen, line 27: Since you use the organizations with a `search`, it is possible that the user has a private profile which will result in a `422` error when querying the search API. You need to provide a fallback for this.

gentlementlegen (changes requested):

src/handlers/shared/start.ts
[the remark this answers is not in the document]
whilefoo: in the SDK it will log even normal Errors so its better to return here

src/plugin.ts
whilefoo: I think the reasoning is that it's needed at multiple places so they just put it in the context

src/utils/issue.ts
```diff
@@ -29,7 +36,7 @@ export async function getAssignedIssues(context: Context, username: string) {
       );
   } catch (err) {
     context.logger.info("Will try re-fetching assigned issues...", { error: err as Error });
-    return getAssignedIssuesFallback(context, username);
+    return getAssignedIssuesFallback(context, username) as Promise<GitHubIssueSearch["items"]>;
```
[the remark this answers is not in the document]
ariesgun: I tried that but in this case, it is needed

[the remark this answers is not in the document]
Keyrxng: update the return type annotation of `getAssignedIssues` to include the `listfor` type that the fallback returns, making it a union of either search or list results

src/types/plugin-input.ts
```diff
@@ -41,6 +41,7 @@ export const pluginSettingsSchema = T.Object(
     taskStaleTimeoutDuration: T.String({ default: "30 Days" }),
     startRequiresWallet: T.Boolean({ default: true }),
     maxConcurrentTasks: maxConcurrentTasks(),
+    checkAssignedIssues: T.String({ default: "org" }),
```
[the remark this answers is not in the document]
Keyrxng: Usually it's better to use 

```ts
const AssignedIssueScope = {
  ORG: "org",
  REPO: "repo",
  NETWORK: "network"
} as const
```

over enums, they are debated to be a bad practice and TS themselves said they regret adding it into the language. It results in less performant JS (because enums don't exist in JS) and can be lost during runtime in libs are not built correctly

They both represent a simple object where `Enum` adds a burden that can be avoided but it's dev preference in this regard as it's a point of debate in the TS world.

src/plugin.ts
```diff
@@ -4,9 +4,46 @@ import { createClient } from "@supabase/supabase-js";
 import { LogReturn, Logs } from "@ubiquity-os/ubiquity-os-logger";
 import { createAdapters } from "./adapters";
 import { userPullRequest, userSelfAssign, userStartStop, userUnassigned } from "./handlers/user-start-stop";
-import { Context, Env, PluginInputs } from "./types";
+import { Context, Env, GitHubIssueSearch, PluginInputs } from "./types";
 import { addCommentToIssue } from "./utils/issue";
 
+async function listOrganizations(context: Context): Promise<string[]> {
```
Keyrxng, line 10: Can you place this in with the handlers or helpers to keep the entry tidy my friend?

ariesgun: sure. will do so

src/types/plugin-input.ts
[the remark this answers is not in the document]
whilefoo: TS enum is better in the way it also creates a type, but in your case you have to do `typeof AssignedIssueScope[keyof typeof AssignedIssueScope]`. I think `const enum` basically generates an object like yours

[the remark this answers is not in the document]
Keyrxng: https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls

const enum is fine here I'm sure because this plugin is not used as a lib but if it was and placed inside a `.ts` file it would be lost at compile time after the file was converted into `.d.ts`. The values will be inlined where they are used.

> Const enums can only use constant enum expressions and unlike regular enums they are completely removed during compilation. Const enum members are inlined at use sites.

> TS enum is better in the way it also creates a type, but in your case you have to do typeof AssignedIssueScope[keyof typeof AssignedIssueScope]

I agree it's a cleaner type definition it just has gotchas with using them so I like to avoid, but dev preference for sure I'm not against them.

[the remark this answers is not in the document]
Keyrxng: [TS Playground](https://www.typescriptlang.org/play/#code/MYewdgzgLgBAghCBLA5mApgEwJKIK7oDKoADujALwwDeAUDDAPIBKA4gFwwBEIATilwA09GMwCiABUacuvdCRBCRAOTEAVAOosA0jIxQA7nwDWXWgF8YAQwgxQkKLXvQY0eQEZK3NyXdmn4C7oYHgAtvCIqBg4+ESk6ADCgVBiIeF0DCysXjz8SgziUjlyCvkwqpo6OfpGvKYWAQ6uUPIATDk+rf60wWERyGhYuBAExCBkqX0ZTGw5fALCBZKMxfKKi+XqWsza1eiGJmbmtLRAA)

this shows how the three enums are compiled

manifest.json
```diff
@@ -46,7 +46,7 @@
           }
         }
       },
-      "checkAssignedIssues": {
+      "assignedIssueScope": {
         "default": "org",
         "type": "string"
       },
```
whilefoo, line 52: are you making this manually? We have a [workflow](https://github.com/ubiquity-os-marketplace/command-start-stop/blob/development/.github/workflows/update-configuration.yml) that automatically updates it, it should include the enum here

whilefoo: let the workflow generate it

src/plugin.ts
gentlementlegen: The spec  states:
> Only check assigned issues within the network, but this should be configurable to either be just within the repo, org, or network

Shouldn't this be configurable? And even if not, putting it in the config also allows to add defaults on decode which would ensure it is properly initialized and types as well.

src/utils/issue.ts
gentlementlegen: When I tested with a user having a private profile, the run crashed on a 422 error. I will test again.

gentlementlegen:
Last QA looks ok: https://github.com/Meniole/command-start-stop/issues/46

src/plugin.ts
whilefoo: maybe I'm missing something, but the config item `assignedIssueScope` can be set to `repo`, `org` or `network`, this `configuration` property gets filled based on `assignedIssueScope` when the plugin starts

whilefoo (approved):

PR #71, opened 2024-10-29
fix: pull-request fetch for private user profiles

### LINKED ISSUE #70: Change `search` queries to GraphQl to avoid permission issues
It is still weird to me that the user privacy affects a search because the profile is public. Can we consider using GQL with issues search instead of the search API? Something like
```gql
query($organization: String!, $author: String!) {
  organization(login: $organization) {
    repositories(first: 100) {
      nodes {
        issues(first: 100, states: OPEN, filterBy: {createdBy: $author}) {
          nodes {
            title
            url
            createdAt
          }
        }
      }
    }
  }
}
```

with

```json
{
  "organization": "ubiquity",
  "author": "sshivaditya2019"
}
```

would achieve the same result. I don't know if that would resolve the issue but it's worth a try.

_Originally posted by @gentlementlegen in https://github.com/ubiquity/work.ubq.fi/issues/119#issuecomment-2440640493_

gentlementlegen:
Resolves #70
QA: https://github.com/Meniole/command-start-stop/issues/41#issuecomment-2444701768 (@ubiquity-ubiquibot has its information private)


gentlementlegen:
@0x4007 On the good news: it works, I successfully retrieve the items! However on the less good news, there are other places where GitHub search is used and they subsequently break as well. I will have to provide a fix for all of them which broadens the scope on this task, on it.

0x4007 (approved):

PR #77, opened 2024-11-04
fix: ensure assignee is checked in userUnassigned handler

### LINKED ISSUE #74: User un-assign should close its linked pull-requests
Disqualify needs to close the pull

_Originally posted by @0x4007 in https://github.com/ubiquity-os-marketplace/daemon-merging/issues/22#issuecomment-2452913872_

gentlementlegen:
Resolves #74
QA: https://github.com/Meniole/command-start-stop/pull/45

### DIFF
--- a/src/handlers/user-start-stop.ts
+++ b/src/handlers/user-start-stop.ts
@@ -34,13 +34,11 @@ export async function userSelfAssign(context: Context<"issues.assigned">): Promi
   const deadline = getDeadline(issue.labels);
 
   // We avoid posting a message if the bot is the actor to avoid double posting
-  if (payload.sender.type === "Bot") {
-    context.logger.debug("Skipping deadline posting message because the sender is a bot.");
-    return { status: HttpStatusCode.NOT_MODIFIED };
-  }
-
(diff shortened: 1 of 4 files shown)


gentlementlegen:
@Keyrxng if you want to take a look.

src/handlers/user-start-stop.ts
```diff
@@ -33,6 +33,12 @@ export async function userSelfAssign(context: Context<"issues.assigned">): Promi
   const { issue } = payload;
   const deadline = getDeadline(issue.labels);
 
+  // We avoid posting a message if the bot is the actor to avoid double posting
+  if (payload.sender.type === "Bot") {
```
Keyrxng, line 37: You could merge with the `if` below since they are the same response

Keyrxng (approved):
Seems to work as intended. 

Previous problem was the sender was used to check for PRs now assignee is used which makes sense

src/handlers/user-start-stop.ts
gentlementlegen: Will do so and merge.

PR #86, opened 2024-11-16
SDK and command interface

whilefoo:
Related to https://github.com/ubiquity-os/ubiquity-os-kernel/pull/186


src/types/plugin-input.ts
```diff
@@ -1,23 +1,19 @@
...
 
-const rolesWithReviewAuthority = T.Array(T.String(), { default: ["COLLABORATOR", "OWNER", "MEMBER", "ADMIN"] });
+export enum Role {
+  OWNER = "OWNER",
+  ADMIN = "ADMIN",
+  MEMBER = "MEMBER",
+  COLLABORATOR = "COLLABORATOR",
+}
+
+const rolesWithReviewAuthority = T.Array(T.Enum(Role), { default: [Role.OWNER, Role.ADMIN, Role.MEMBER, Role.COLLABORATOR], uniqueItems: true });
```
whilefoo, line 16: With Typebox v33+, the array gets filled with default values even if it's already filled, for example the test for default supplies `[Role.OWNER, Role.ADMIN, Role.MEMBER]` and after `Value.Default` is called, the array is `[Role.OWNER, Role.ADMIN, Role.MEMBER, Role.COLLABORATOR]`.
This behaviour is not logical to me, but I can't find a way to make it behave like before other than downgrading to v32

whilefoo:
@gentlementlegen can't request you for review

src/types/plugin-input.ts
gentlementlegen: That would explain the issue I just had here: https://github.com/ubiquity-os-marketplace/text-conversation-rewards/pull/186

that's quite annoying indeed.

gentlementlegen: @whilefoo Seems that it got taken care of! https://github.com/sinclairzx81/typebox/pull/1083

whilefoo: That's awesome and we didn't even make an issue :D

whilefoo:
Did you link to the new SDK using `bun link @ubiquity-os/plugin-sdk`? 
I'm thinking if manifest and other types should be inside the kernel and the SDK imports the kernel, it makes more sense to make kernel the source of truth so to say.

> They are an array when they should be an object, why is that?

Honestly I thought it's easier as an array to just put these objects into OpenAI functions but later realized it's neither easier nor harder, I can change back to the object if you think it's better

gentlementlegen:
@whilefoo The plus side of the object is that it prevents duplicate keys.

Yes I usually link with `--watch` so I don't have to think about it. We can get the code back to the kernel if you think that is beneficial, but as we discussed before we need to absolutely make sure there is no circular reference of any import otherwise that will break the compilation.

I had an awful time splitting the SDK itself, because of the following:
- if everything is compiled within `index.d.ts` then every lib is included which is useless and can lead to circular imports or incompatible packages
- splitting it with subpath is not possible except if using `node16` option at least
- `node16` is really bad because it forces silly things like adding `.js` at the end of the import paths, and breaks some libs
- there is the `bundler` option that mitigates that but doesn't solve everything
- the trick I found was to use the future imports, that leverage part of the node16 and upwards features while being compatible with the plain `node` as seen [here](https://github.com/ubiquity-os/plugin-sdk/blob/development/package.json#L14)

Smashed my head on the desk a whole day to fix this.

whilefoo:
> Yes I usually link with `--watch` so I don't have to think about it. We can get the code back to the kernel if you think that is beneficial, but as we discussed before we need to absolutely make sure there is no circular reference of any import otherwise that will break the compilation.

I assume you built the SDK? It seems like you're using the old SDK version, but I'm not sure what's wrong, I've tried again on my end and it works

> * the trick I found was to use the future imports, that leverage part of the node16 and upwards features while being compatible with the plain `node` as seen [here](https://github.com/ubiquity-os/plugin-sdk/blob/development/package.json#L14)

Ohh, that's why I couldn't import `octokit` subpath that I added, I didn't see that I also need to add `typesVersion`. I also tried to switching to different module resolutions but nothing worked

gentlementlegen:
@whilefoo yes that's another reason why I switched to bun in the conversation rewards because it is easier to link with the same package manager. Usually using bun instead of yarn temporarily to install packages has not been a problem.

gentlementlegen (approved):

PR #111, opened 2024-12-18
feat: deploy url

gentlementlegen:
Relates to https://github.com/ubiquity-os/plugin-template/issues/33


rndquu (approved):

gentlementlegen:
Seems `ncc` failed to compile this one: https://github.com/Meniole/command-start-stop/actions/runs/12391955983/job/34590118129
Fixing.

.github/workflows/worker-deploy.yml
```diff
@@ -54,6 +61,46 @@ jobs:
...
+          bun prettier --write .
+
+      - name: Commit file
+        uses: swinton/commit@v2.x
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+        with:
+          files: |
+            manifest.json
+          commit-message: "chore: [skip ci] update manifest.json url"
```
whilefoo, line 101: nice, I didn't know you can skip CI with that

whilefoo (approved):

.github/workflows/worker-deploy.yml
gentlementlegen: Some hidden features of GitHub
https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs

gentlementlegen:
It seems that after `bun` got upgraded to `1.40`, `ncc` started to fail complaining about types. My theory is that `bun 1.38` was not resolving package dependencies properly and used `typescript 5.5.4` (which compiles fine) instead of `5.6` which fails (also tested with `5.7`, same result). Downgraded TypeScript to `5.5.4` and it compiles just fine: https://github.com/Meniole/command-start-stop/actions/runs/12392695131

PR #115, opened 2024-12-28
fix: linked pull-requests on tasks that cannot be started get closed

### LINKED ISSUE #102: Comment fails to be posted on linked pull-request that cannot be started
We currently have restrictions on the level of tasks we are able to start (at the moment at least priority 3). However, users can still link a pull-request to that issue if they want to. What happens is that the plugin properly recognizes that the task cannot be started ("`⚠ This task does not reflect a business priority at the moment. You may start tasks with one of the following labels: Priority: 3 (High), Priority: 4 (Urgent), Priority: 5 (Emergency)`") but fails to post that comment because the starting event comes from the pull-request and not the issue, thus failing with `› Cannot post comment because issue is not found in the payload`. This comment should be posted otherwise users could think that the plugin is not working properly.

## What is to be done
- properly post the comment that the user cannot start this task
- add related tests

## Example of failed run
https://github.com/ubiquity-os-marketplace/daemon-disqualifier/issues/60#issuecomment-2525053118

gentlementlegen:
Resolves #102
Depends on https://github.com/ubiquity-os/plugin-sdk/pull/42 to properly post errors
QA: https://github.com/Meniole/command-start-stop/pull/53


gentlementlegen:
Need https://github.com/ubiquity-os/plugin-sdk/pull/42 to be resolved first.

whilefoo (approved):

gentlementlegen:
@whilefoo updated the plugin to use the SDK to post messages with the new version we released yesterday. Re-QAd here: https://github.com/Meniole/command-start-stop/issues/54#issuecomment-2574376735

PR #116, opened 2024-12-29
fix: member role inside of given repository checks

### LINKED ISSUE #114: Check collaborator status for users
image

@gentlementlegen he's a collaborator this error should not have been thrown

_Originally posted by @0x4007 in https://github.com/ubiquity/business-development/issues/106#issuecomment-2563943154_
            

## What happened

The user was recognized as `member` instead of collaborator, thus not being able to start the task.

## What was expected

The user should have been recognized as `collaborator` and start the task.

gentlementlegen:
Resolves #114
QA: https://github.com/Meniole/command-start-stop/issues/52#issuecomment-2568788596 (ubiquibot is an outside collaborator with write permissions)


0x4007:
@gentlementlegen  will you be able to prioritize this？

gentlementlegen:
@0x4007 Yes waiting on your input about https://github.com/ubiquity-os-marketplace/command-start-stop/issues/114#issuecomment-2564665330

whilefoo (approved):
I'm fine with the changes, my only concern is that we will have to declare max concurrent tasks for every permission (write, read, ...) and it might be confusing

gentlementlegen:
@whilefoo maybe we could simplify to just `collaborator` and `contributor`, and then map a group of roles to these two? Admin would just have no limits.

0x4007:
That sounds like a good plan.

PR #119, opened 2025-01-04
fix: simplified member role in configuration

### LINKED ISSUE #118: Simplify role limits by grouping all the roles under "collaborator" and "contributor"
@whilefoo maybe we could simplify to just `collaborator` and `contributor`, and then map a group of roles to these two? Admin would just have no limits. I can do that in a secondary pull-request. @0x4007 rfc

_Originally posted by @gentlementlegen in https://github.com/ubiquity-os-marketplace/command-start-stop/issues/116#issuecomment-2569982989_

To simplify the roles and avoid a long and complex configuration, we should regroup roles under `collaborator` (anyone with write access or from the organization) and `contributor` (everyone else). Admins should have no limits.

gentlementlegen:
Resolves #118
QA: https://github.com/Meniole/command-start-stop/issues/54


gentlementlegen:
/ask what are the changes in this pull-request about?

src/types/plugin-input.ts
```diff
@@ -23,47 +28,24 @@ const rolesWithReviewAuthority = T.Array(T.Enum(Role), {
...
 const requiredLabel = T.Object({
   name: T.String({ description: "The name of the required labels to start the task." }),
-  roles: T.Array(roles, {
+  allowedRoles: T.Array(roles, {
     description: "The list of allowed roles to start the task with the given label.",
     uniqueItems: true,
-    default: ["admin", "member", "collaborator", "contributor", "owner", "billing_manager", "write", "read"],
+    default: ["collaborator", "contributor"],
+    examples: [["collaborator", "contributor"]],
   }),
```
whilefoo, line 49: since admin is not in settings anymore how would one restrict a label to just admins?

gentlementlegen: I removed the defaults, so that if you omit both values only admins will be able to start the task. QA: https://github.com/Meniole/command-start-stop/issues/54#issuecomment-2585573332

whilefoo: Maybe I'm missing something but for example if I set empty array and I'm a collaborator then this [if](https://github.com/ubiquity-os-marketplace/command-start-stop/pull/119/files#diff-47224fac5bab974df0f6fb9704d6fbb0f06375b3e98285797826104537bda3ccR34-R38) would evaluate to `false` so it would let me start the task

gentlementlegen: You are correct, I didn't anticipate the different logic if we have "contributor only" (which I think would never happen but it is handled now), and updated the message for admin only tasks (same thing). QAd: https://github.com/Meniole/command-start-stop/issues/55

Also now admin also bypasses required labels which I think makes more sense.

whilefoo (approved):

gentlementlegen:
Updated configuration for this as well. I wish we had the frontend for this because I am sure I missed some repos with custom configs.

PR #125, opened 2025-01-14
fix: validate presence of price label in issue before checking requir…

### LINKED ISSUE #99: Wrong Order of Start Errors
- Ideally should aggregate all the potential problems for starting and explain to the user. 
- At least the first error should say that there is no price label so this can't be started.

```diff
! This task does not reflect a business priority at the moment. You may start tasks with one of the following labels: Priority: 3 (High), Priority: 4 (Urgent), Priority: 5 (Emergency)
```

_Originally posted by @ubiquity-os[bot] in https://github.com/ubiquity/business-development/issues/93

koya0:
…ements

Resolves #99


koya0:
@0x4007 could you please review

0x4007:
Can you link the thread instead of screenshots? Better to QA when we can use it directly.

src/handlers/shared/start.ts
```diff
@@ -67,6 +67,13 @@ export async function start(
     throw logger.error(`Skipping '/start' since there is no sender in the context.`);
   }
 
+  const labels = issue.labels ?? [];
+  const priceLabel = labels.find((label: Label) => label.name.startsWith("Price: "));
+
+  if (!priceLabel) {
+    throw logger.error("No price label is set to calculate the duration", { issueNumber: issue.number });
```
0x4007, line 74: What about error aggregation?

koya0:
> Can you link the thread instead of screenshots? Better to QA when we can use it directly.

i ended up deleting those issues, but i created other ones:
before - https://github.com/koya0/.ubiquity-os/issues/24
after - https://github.com/koya0/.ubiquity-os/issues/25

src/handlers/shared/start.ts
koya0: as gentlemen said it was complex, i just changed the order, but I can make the aggregation

koya0: you think its okay to aggregate only the price label with the errors from checkRequirements? or you mean ALL the erros?

0x4007: The idea is that I want to avoid the situation where a user steps through every error and loses patience. 

For example:

- you must set your wallet before starting the issue

And then they set their wallet and try start again

- this is not a business priority please start tasks with level 3 priority or higher 

Etc

koya0: i ended up aggregating them, you can take a look at https://github.com/koya0/.ubiquity-os/issues/26

koya0: other errors that Ithought maybe could be aggregated were "Only collaborators can be assigned to this issue." and "You have reached your max task limit. Please close out some tasks before assigning new ones."

koya0: > The idea is that I want to avoid the situation where a user steps through every error and loses patience. 
> 
> For example:
> 
> - you must set your wallet before starting the issue
> 
> And then they set their wallet and try start again
> 
> - this is not a business priority please start tasks with level 3 priority or higher 
> 
> I feel like there were other temporary errors I can't recall at the moment. I figured aggregation would prevent this from happening

Got it! Its a really good idea. I think the only problem is that it would be a lil complex to aggregate errors from differents plugins

0x4007: I think just worry about the errors in this plugin for now.

0x4007 (approved):
Aggregated errors QA looks good! 

https://github.com/ubiquity-os-marketplace/command-start-stop/pull/125#discussion_r1920911342

gentlementlegen:
QA seems fine: https://github.com/Meniole/command-start-stop/issues/59

Please bump `@ubiquity-os/plugin-sdk` to `2.0.2`

Also, it appears that a task with no time estimate can be started if it has a price and a priority, I don't know if that's expected.

0x4007:
> QA seems fine: [Meniole#59](https://github.com/Meniole/command-start-stop/issues/59)
> 
> Please bump `@ubiquity-os/plugin-sdk` to `2.0.2`
> 
> Also, it appears that a task with no time estimate can be started if it has a price and a priority, I don't know if that's expected.

I suppose that technically speaking all we need is a `Price` label.

gentlementlegen:
/ask Could this pull-request be rewritten using `AggregateError` class to compound all the errors together?

koya0:
> Even though it works fine, I think that returning a string this way is not really clear, programming side. Why not using `AggregateError` class and push all the errors there using the logger, and then throw the whole errors altogether which would be handled directly by the SDK? That would also properly break the flow of the program.

is this class implemented by `@ubiquity-os/plugin-sdk`?

gentlementlegen:
You should be able to import it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError

koya0:
I ended up not using `AggregateError`, because what I would use of it would be `AggregateError.errors`, which is the same of startErrors array. You can check the new comment format here: https://github.com/koya0/.ubiquity-os/issues/26#issuecomment-2608585881

gentlementlegen:
The formatting is ok. The thing I wanted to avoid is

https://github.com/ubiquity-os-marketplace/command-start-stop/pull/125/commits/cb32e3fc58e1a3b695fee9647292dff7d4a273a4#diff-47224fac5bab974df0f6fb9704d6fbb0f06375b3e98285797826104537bda3ccR89

where I would prefer to only throw one error and the SDK would handle the display. Because by doing so, what is the linked error stack? I can't edit the comment so I can't see it.

0x4007:
I think don't use `new Error()` because then it prefixes all the messages with "Error: " which is redundant and doesn't look good. Obviously the whole thing is an error.

koya0:
> I think don't use `new Error()` because then it prefixes all the messages with "Error: " which is redundant and doesn't look good. Obviously the whole thing is an error.

if I dont use `new Error()` I guess would have to return a string, which gentlemen said it was not a good option. but this "Error: " prefix can be removed by logging only the .message of the error

gentlementlegen:
@koya0 could you fix the jest tests? Then I'll be able to merge this.

koya0:
> @koya0 could you fix the jest tests? Then I'll be able to merge this.

done, sorry for the delay 😔

PR #126, opened 2025-01-15
fix: condition for branch target changed to be either ref or workflow ref

### LINKED ISSUE #124: Check deployment target branch
> I think I figured out why, basically the main version is not up to date. This message doesn't exist anymore. 

 _Originally posted by @gentlementlegen in [#111](https://github.com/ubiquity/business-development/issues/111#issuecomment-2589563669)_

The deployment target branch seems to always be the default instead of the branch where the build occurred.

gentlementlegen:
Resolves #124
QA:
- [development branch build](https://github.com/Meniole/command-start-stop/actions/runs/12791736846)
- [targeted branch build](https://github.com/Meniole/command-start-stop/actions/runs/12791780600)

### DIFF
--- a/.github/workflows/worker-delete.yml
+++ b/.github/workflows/worker-delete.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Get Deleted Branch Name
         id: get_branch
         run: |
-          branch_name=$(echo '${{ github.event.ref }}' | sed 's#refs/heads/##' | sed 's#[^a-zA-Z0-9]#-#g')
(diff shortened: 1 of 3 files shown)


whilefoo (commented):
If I understand correctly, because this workflow gets triggered after another workflow run, `github.ref` will always be the default branch of the repo instead of the branch that Update Configuration and Build` ran on, so you can get that branch in `workflow_run.head_branch`?

gentlementlegen:
@whilefoo correct, please refer to https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run

`github.ref` would have the proper value if the build is manually triggered, and defaults to the default branch when following a `workflow_run`.

whilefoo (approved):

PR #133, opened 2025-01-29
fix: linking a pull-request to un un-priced task does not get close

### LINKED ISSUE #129: Do not close pull-requests linked to an issue with no price tag
I think it basically should be disabled in respect to automations for normal GitHub issues without pricing. Then it won't get in the way for partners easing into using the system

_Originally posted by @0x4007 in https://github.com/ubiquity-os/ubiquity-os-kernel/issues/234#issuecomment-2599513665_
            

## What happened

The pull-request that was linked to the issue got closed because the related issue had no price tag set.

## What was expected

The pull-request should not get closed when the price label is missing.

## How to reproduce

- create an issue with no price tag, only time estimate for example
- link a pull-request that would close it (Resolves #)
- notice that the pull-request gets closed by the bot

gentlementlegen:
Resolves #129
QA: https://github.com/Meniole/command-start-stop/issues/60


src/handlers/shared/start.ts
```diff
@@ -71,16 +73,17 @@ export async function start(
...
 
   const startErrors: Error[] = [];
 
-  if (!priceLabel) {
-    const errorText = "No price label is set to calculate the duration";
-    logger.error(errorText, { issueNumber: issue.number });
-    startErrors.push(new Error(errorText));
+  // Collaborators and admins can start un-priced tasks
+  if (!priceLabel && userRole === "contributor") {
+    startErrors.push(new Error(logger.error("No price label is set to calculate the duration", { issueNumber: issue.number }).logMessage.raw));
```
0x4007, line 83: Why do you need to wrap the logger call with an error that seems wrong

gentlementlegen: These changes were made when you requested having the errors thrown together. I just one lined it.

0x4007: Still seems wrong. Why not remove new Error

gentlementlegen: Because `startErrors` in an array of error, `logger.error` is a `LogReturn`, I don't know if `AggregateError` can handle non Error objects, I can try but would be weird to store objects unrelated to errors there.

whilefoo: This is why I proposed long time ago that `LogReturn` should be extending `Error` class so you can use it anywhere that accepts `Error` type

tests/start.test.ts
```diff
@@ -130,7 +132,81 @@ describe("Collaborator tests", () => {
...
+          //@ts-expect-error partial mock of the endpoint
+          getMembershipForUser: jest.fn(() => ({ data: { role: "member" } })),
+        },
+      },
+    };
+    await expect(startStopTask(context)).rejects.toMatchObject({
+      logMessage: {
+        raw: expect.stringContaining("This task does not reflect a business priority at the moment"),
+      },
+    });
```
whilefoo, line 210: this test can be deceiving because there could be both errors (no price label + no business priority) so you wouldn't actually know that it's not working

src/handlers/shared/start.ts
0x4007: I think you can make a spec and take initiative on it if it's a problem.

gentlementlegen: Semantically feels weird to me that a logger returns error based class though. I don't see the problem in wrapping this in an error, it is just a string that I inlined. I can make it multiple lines again if that's bothering you.

src/handlers/user-start-stop.ts
```diff
@@ -107,9 +107,9 @@ export async function userPullRequest(context: Context<"pull_request.opened" | "
         try {
           return await start(context, issueWithComment, pull_request.user ?? payload.sender, []);
         } catch (error) {
-          context.logger.info("The task could not be started, closing linked pull-request.", { pull_request });
           await closePullRequest(context, { number: pull_request.number });
-          throw error;
+          // Makes sure to concatenate error messages on AggregateError for proper display
+          throw error instanceof AggregateError ? context.logger.error(error.errors.map(String).join("\n"), { error }) : error;
```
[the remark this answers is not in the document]
gentlementlegen: https://github.com/ubiquity-os/plugin-sdk/issues/64

tests/start.test.ts
gentlementlegen: Changed to strict equal: https://github.com/ubiquity-os-marketplace/command-start-stop/pull/133/commits/03a9bdb694ac81a1202658fbdd56583655d1166d

src/handlers/shared/start.ts
0x4007: It's just weird to me that you have to store error objects in the array. 

All the other logger classes just work with strings.

tests/main.test.ts
```diff
@@ -152,11 +148,7 @@ describe("User start/stop", () => {
     expect(content).toEqual("Linked pull-requests closed.");
     const logs = infoSpy.mock.calls.flat();
     expect(logs[0]).toMatch(/Opened prs/);
-    expect(cleanLogString(logs[3])).toMatch(
-      cleanLogString(
-        " › ```diff# These linked pull requests are closed:  http://github.com/ubiquity/test-repo/pull/2  http://github.com/ubiquity/test-repo/pull/3"
-      )
-    );
+    expect(cleanLogString(logs[3])).toMatch(cleanLogString("›Closinglinkedpull-request."));
```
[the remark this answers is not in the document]
gentlementlegen: Upgrading the logs package removed the diff display in favor of the `CAUTION` display. I think the cleanLogString removes spaces which is why it looks like this, and I didn't want to change all the tests, but I can in a separate PR.

whilefoo (approved):

PR #134, opened 2025-01-29
feat: correct context and skip issues outside of the org

### LINKED ISSUE #131: Assign user to the proper linked issue
> There's a bug from linking from here
> 
> https://github.com/ubiquity-os-marketplace/text-conversation-rewards/pull/253 

 _Originally posted by @0x4007 in [#46](https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/46#issuecomment-2620290902)_

## What happened

The user got assigned to an issue with a similar `number` but on the wrong repository.

## What was expected

The user should get assigned to the linked issue.

whilefoo:
Resolves #131

The issue is that `start` function was getting the `context` associated with the PR's repo but the issue was from another repo. I fixed it by creating a new context from the old context and replacing `issue` and `repository` from the API, but this feels more like a hack because the `config` will be from the original repo.

Another problem is that the PR's repo and issue's repo have to be in the same organization because the plugin only has installation token for the org and it doesn't have write permissions in other orgs. For example I made a PR in `ubiquity-os-marketplace/text-conversation-rewards` and linking to `ubiquity-os/plugins-wishlist`, this won't work because of the permissions.
A solution would be that we somehow move the logic so that it happens in the issue instead of the PR, but I couldn't find any webhook that would be triggered when an issue got a PR linked, cross-referenced or something similar to that so that we could trigger it based on that.


gentlementlegen:
What about creating a new Octokit instance with the proper credentials for the other organization from this plugin? It is possible because we have the APP_ID and the APP_PRIVATE_KEY, and this has been used in other plugins to solve authentication issue. Maybe it would be worth having this function exposed from the SDK.

whilefoo:
Enabled support for cross-org with `APP_ID` and `APP_PRIVATE_KEY`

cross-org QA: [PR](https://github.com/ubiquity-whilefoo/testing/pull/8) and [issue](https://github.com/ubiquibot-whilefoo-testing/testing/issues/8#issuecomment-2625285023)

src/handlers/user-start-stop.ts
```diff
@@ -1,11 +1,13 @@
...
-import { Context, isIssueCommentEvent, Label } from "../types";
+import { Context, isIssueCommentEvent } from "../types";
 import { QUERY_CLOSING_ISSUE_REFERENCES } from "../utils/get-closing-issue-references";
 import { closePullRequest, closePullRequestForAnIssue, getOwnerRepoFromHtmlUrl } from "../utils/issue";
 import { HttpStatusCode, Result } from "./result-types";
 import { getDeadline } from "./shared/generate-assignment-comment";
 import { start } from "./shared/start";
 import { stop } from "./shared/stop";
+import { customOctokit } from "@ubiquity-os/plugin-sdk/octokit";
+import { createAppAuth } from "@octokit/auth-app";
```
gentlementlegen, line 10: Nitpicking, but I think the imports are not in alphabetical order (and that creates more merge conflicts).

0x4007: I have some vscode plugin for this. I'm not on computer so I don't know off hand but one of the popular typescript ones I'm guessing like typescript hero. 

Anyways I'm sure you can find some plugin to do this automatically for you.

0x4007:
QA isn't 100% clear to me I left a comment on QA PR

src/handlers/user-start-stop.ts
gentlementlegen: My IDE does it automatically (Webstorm). VSCode also has an option for that as far as I remember. Advantage of alphabetical sorting is that on merge import are much less likely to cause merge conflicts because they are always arranged in the same order.

0x4007 (approved):

gentlementlegen:
Tested it, not sure if I did something wrong but the issues doesn't appear to be fixed:
- created a pull-request: https://github.com/Meniole/command-start-stop/pull/63
- linked to another organization: https://github.com/mentlegen-2/project/issues/10
- got assigned to the wrong issue: https://github.com/Meniole/command-start-stop/issues/10

gentlementlegen (approved):

PR #155, opened 2025-03-19
chore: export checkRequirements function

zugdev:
can we please export this function, would be useful for UI task gating: https://github.com/ubiquity/work.ubq.fi/issues/206

### DIFF
--- a/src/handlers/shared/start.ts
+++ b/src/handlers/shared/start.ts
@@ -9,7 +9,7 @@ import { getTransformedRole, getUserRoleAndTaskLimit } from "./get-user-task-lim
 import structuredMetadata from "./structured-metadata";
 import { assignTableComment } from "./table";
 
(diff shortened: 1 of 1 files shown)


zugdev:
@gentlementlegen @0x4007

0x4007:
Are you importing as a submodule? That would be epic. @gentlementlegen said its most likely not possible due to Octokit having nodejs dependencies.

zugdev:
> Are you importing as a submodule? That would be epic. @gentlementlegen said its most likely not possible due to Octokit having nodejs dependencies. But I feel with `ncc` anything is possible. I just didn't have time to personally look into this. 

trying to do that! I agree it would be sick

> Also in the future for auditing and documentation purposes, it would be ideal if you can reference the original task you are working on in your pull body. Not resolves but "reference"

gotcha

PR #161, opened 2025-04-28
fix: deploy to azure

gentlementlegen:
QA: https://github.com/Meniole/command-start-stop/issues/67#issuecomment-2835941505

Moving this to Azure because in vercel we will always exceed the 100GB data quickly.


.github/workflows/azure-deploy.yml
```diff
@@ -0,0 +1,148 @@
...
+# More GitHub Actions for Azure: https://github.com/Azure/actions
+
+name: Build and deploy Node.js project to Azure Function App - ubiquity-os-command-start-stop
+
+on:
+  push:
+    branches:
+      - development
+      - main
+  workflow_dispatch:
```
whilefoo, line 11: shouldn't this be run after `Update Configuration and Build` workflow like with CF workers?

tsup.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/functions/*.ts"],
+  splitting: false,
+  sourcemap: false,
+  clean: true,
+  format: ["esm"],
+  minify: true,
+});
```
whilefoo, line 10: why don't we use `ncc` like usually?

.github/workflows/azure-deploy.yml
```diff
@@ -0,0 +1,148 @@
...
+          tenant-id: ${{ secrets.AZUREAPPSERVICE_TENANTID }}
+          subscription-id: ${{ secrets.AZUREAPPSERVICE_SUBSCRIPTIONID }}
+
+      - name: "Run Azure Functions Action"
+        uses: Azure/functions-action@v1
+        id: fa
+        with:
+          app-name: "${{ env.FUNCTION_NAME }}"
+          slot-name: ${{ github.ref == 'refs/heads/main' && 'Production' || 'development' }}
+          package: ${{ env.AZURE_FUNCTIONAPP_PACKAGE_PATH }}
```
whilefoo, line 148: do we get deployment URL like with CF workers?
and I'm guessing Azure doesn't have preview URLs

.github/workflows/azure-deploy.yml
```diff
@@ -0,0 +1,148 @@
...
+
+      - name: Azure CLI script to upload environment
+        uses: azure/cli@v2
+        with:
+          azcliversion: latest
+          inlineScript: |
+            az account show
+            az functionapp config appsettings set \
+              --name ${{ env.FUNCTION_NAME }} \
+              --resource-group ubiquity-os_group \
```
whilefoo, line 114: resource group name is chosen by the user, right? so I think we should define it as an env

.github/workflows/azure-deploy.yml
gentlementlegen: It depends. In Azure, there is something called "slots" which I use here (one for development and one for production) but within the current plan I can only have two active slots so I can't get a preview for each branch sadly.

.github/workflows/azure-deploy.yml
gentlementlegen: Yes it is the created resource group name within Azure, I will change it to be in the env then.

tsup.config.ts
gentlementlegen: I did this because `tsup` has a watch mode which is handy during local development, but yes I can switch it to `ncc` most likely. This being said tsup is much faster as well and easier to configure.

.github/workflows/azure-deploy.yml
gentlementlegen: We could, but since we only support these two branches we would have lots of failed runs. Plus, it has its own build step right now.

.github/workflows/azure-deploy.yml
gentlementlegen: https://github.com/ubiquity-os-marketplace/command-start-stop/pull/161/commits/417a5336bdba10ebe19698811bd891ca60cc2359

tsup.config.ts
gentlementlegen: image

I tried to use `ncc` for build but I could never get it to work. I know for a fact that `ncc` includes too many things in its builds like the tests, and I never could get rid of this behavior. In Azure I had to be careful about what I include in the build because each function is considered as an entry point.

https://github.com/ubiquity-os-marketplace/command-start-stop/blob/2f0d88c296b21028673c727e32a2646b8de537e2/.github/workflows/azure-deploy.yml#L86-L87

whilefoo:
Hmm I don't remember clicking merge...anyway I tried deploying to Azure and encountered this [issue](https://github.com/ubiquity-os-marketplace/command-start-stop/actions/runs/14865634737/job/41747618446). Also I'm curious why the workflow doesn't use client secret or something else, it only needs IDs so I don't understand how does it authenticate

gentlementlegen:
@whilefoo it seems your merged it through the PR https://github.com/ubiquity-os-marketplace/command-start-stop/pull/162 (did you target the wrong repo for testing?)

I did not set these variables in the env yet I think, let me check and retry a deployment.

whilefoo:
> @whilefoo it seems your merged it through the PR #162 (did you target the wrong repo for testing?)

Ah I didn't even realize, I thought I merged to my development branch

I created Managed Identity and copied the client id but I think I'm still missing something: https://github.com/ubiquity-whilefoo/command-start-stop/actions/runs/14878918097/job/41783034879

gentlementlegen:
It doesn't matter which one runs first because both will trigger a reload of the function.

It seems that you didn't created federated credentials, cf my screenshot:

image


You have to allow the environment (not the branch) for it to work.

whilefoo:
ok got the federated creds working but now I get [`No subscriptions found for...`](https://github.com/ubiquity-whilefoo/command-start-stop/actions/runs/14878918097/job/41901307116). Any ideas?

gentlementlegen:
My guess is that you didn't add that IAM user as a contributor of the Function
image

whilefoo:
Ah now the login works but my function name is different because it must be unique so I think that's why it failed, so we must also put function name in the secrets/env
https://github.com/ubiquity-whilefoo/command-start-stop/actions/runs/14878918097/job/41985919954

gentlementlegen:
Ha yes makes sense. Instead of having it in the secrets we could also determine this from the repo name.

whilefoo:
I manually changed function name but I guess I'm still missing something https://github.com/ubiquity-whilefoo/command-start-stop/actions/runs/14956787670/job/42013638166

gentlementlegen:
Never had this issue but my guess would be that no storage was set when you created the function, which is why this error is shown. Also I usually tend to use windows instead of linux so I can have access to live logs.

PR #165, opened 2025-05-27
feat: deno

gentlementlegen:
Resolves https://github.com/ubiquity-os/ubiquity-os-kernel/issues/275

Requires `DENO_DEPLOY_TOKEN` to be set in the Action env.

### DIFF
--- a/.github/workflows/deno-deploy.yml
+++ b/.github/workflows/deno-deploy.yml
@@ -20,7 +20,7 @@ jobs:
           branch_name=$(echo '${{ github.event.ref || github.event.workflow_run.head_branch || github.ref }}' | sed 's#refs/heads/##')
           echo "ACTION_REF=${GITHUB_REPOSITORY}@${branch_name}" >> $GITHUB_ENV
 
-      - uses: Meniole/deno-deploy@main
+      - uses: ubiquity-os/deno-deploy@main
         env:
           APP_ID: ${{ secrets.APP_ID }}
           APP_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}


src/handlers/shared/generate-assignment-comment.ts
```diff
@@ -1,4 +1,4 @@
-import { Context } from "../../types";
+import { Context } from "../../types/index";
```
0x4007, line 1: Is index really necessary

src/types/payload.ts
```diff
@@ -7,6 +7,7 @@ export type TimelineEventResponse = RestEndpointMethodTypes["issues"]["listEvent
 export type TimelineEvents = RestEndpointMethodTypes["issues"]["listEventsForTimeline"]["response"]["data"][0];
 export type Assignee = Issue["assignee"];
 export type GitHubIssueSearch = RestEndpointMethodTypes["search"]["issuesAndPullRequests"]["response"]["data"];
+export type PrState = "open" | "closed" | "all" | undefined;
```
0x4007, line 10: Undefined seems wrong

src/handlers/shared/generate-assignment-comment.ts
gentlementlegen: Yes otherwise appending ".ts" extensions to files would not work. It could probably be handled but that would make the complexity of the action much higher. Could be done in a V2 eventually.

src/types/payload.ts
gentlementlegen: I took directly the type from "@octokit/types" amd dropped it here

0x4007 (approved):

PR #181, opened 2025-09-18
fix: normalize user role handling

### LINKED ISSUE #179: Collaborator assignment
## What happened

I tried to assign myself with /start in this issue https://github.com/ubiquity-os-marketplace/personal-agent/issues/7 and I got the message saying `External contributors are not eligible for rewards at this time. We are preserving resources for core team only.`

Metadata:
```json
{
  "userRole": "collaborator",
  "price": 9,
  "userAllowedMaxPrice": -1,
  "issueNumber": 7,
  "caller": "start"
}
```
Metadata shows that I was correctly identified as a collaborator but the max allowed price is -1 which is not correct because the config is set to Infinity for collaborators:
```yaml
taskAccessControl:
  usdPriceMax:
    contributor: -1
    collaborator: Infinity
```

## Expected outcome

I should've been assigned because I'm member of the org / collaborator.

gentlementlegen:
Resolves #179

To be honest, I could not reproduce the issue despite having the same config and the same user access level. I did notice a possible discrepancy with the roles, but can't confirm that is a valid fix.

Also, the logs do not exist anymore so I don't have the details.


0x4007 (commented):
We can try testing in prod. Best to push this branch on this repo and then target it directly and have @whilefoo try again 

In this case it might make sense to split the assignee between you two so that it can be tested properly to diagnose.

gentlementlegen:
I'll try to use the exact same version of the plugin and the configuration at that time, and see. 

image

image


Copied what you had on your screenshots.

whilefoo:
I've just tested in the [same org but different repo](https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/433#issuecomment-3320175912) and [same org and same repo](https://github.com/ubiquity-os-marketplace/personal-agent/issues/11#issuecomment-3320186807) and both times it worked.
Maybe it was a temporary API error?

PR #193, opened 2025-12-03
When labels are missing, it shows a warning instead of an error

### LINKED ISSUE #191: Change caution to warning
> [!CAUTION]
> No labels are set.

_Originally posted by @ubiquity-os-beta[bot] in https://github.com/ubiquity-os/action-deploy-plugin/issues/24#issuecomment-3572095388_

The `caution` message for unset labels should be changed to `warning`.

gentlementlegen:
Resolves #191

QA: https://github.com/Meniole/command-start-stop/issues/104#issuecomment-3605614285

Changed the throws to a comment post, so we can keep the original level of the posted message instead of always throwing.


src/handlers/start/helpers/error-messages.ts
```diff
@@ -107,5 +107,6 @@ ${issues}
     throw new AggregateError(errorMessages.map((e) => new Error(e.logMessage.raw)));
   }
 
-  throw errorMessages[0];
+  await context.commentHandler.postComment(context, errorMessages[0]);
+  return { content: errorMessages[0].logMessage.raw, status: HttpStatusCode.BAD_REQUEST };
```
0x4007: It appears that the logic is different here.

gentlementlegen: Logic of not throwing? The other one is thrown because it is a list of errors, and here a single one

